PR #24, opened 2021-09-27
Centralizes the logic for retrieving context information

elliot-smith:
*Issue #, if available:*

*Description of changes:*
Currently, we retrieve the context information in several locations using different error messages. This change allows us to centralize the error messaging between the different components.

### DIFF
--- a/packages/cli/internal/pkg/cli/spec/project.go
+++ b/packages/cli/internal/pkg/cli/spec/project.go
@@ -1,5 +1,7 @@
 package spec
 
+import "fmt"
+
 const LatestVersion = 1
 
 type Project struct {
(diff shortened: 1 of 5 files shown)


packages/cli/internal/pkg/cli/spec/project.go
```diff
@@ -9,3 +11,12 @@ type Project struct {
 	Data          []Data              `yaml:"data,omitempty"`
 	Contexts      map[string]Context  `yaml:"contexts,omitempty"`
 }
+
+func GetContext(projectSpec Project, contextName string) (Context, error) {
```
tneely, line 15: Rather than have this be a standalone function, you simplify it by defining it as a method on the `Project` struct:

```
func (p Project) GetContext(contextName string) (Context, error) {
  contextSpec, ok := p.Contexts[contextName]
  //...
}
```

elliot-smith: Fixed

packages/cli/internal/pkg/cli/spec/project.go
```diff
@@ -9,3 +11,12 @@ type Project struct {
 	Data          []Data              `yaml:"data,omitempty"`
 	Contexts      map[string]Context  `yaml:"contexts,omitempty"`
 }
+
+func (projectSpec *Project) GetContext(contextName string) (Context, error) {
```
markjschreiber, line 15: Nit, because Go doesn't allow overloaded methods, possibly a better name would be GetContextForName

markjschreiber (approved):

packages/cli/internal/pkg/cli/spec/project.go
elliot-smith: I thought about that but the current standard in the project is to access the fields directly instead of using getters/setters.

tneely (approved):

PR #37, opened 2021-09-29
Adds workflow type to the command use example for project init

elliot-smith:
*Description of changes:*
Updates the Use part of the project init command. I also checked the documentation but we include workflow type there which is why it isn't updated there.

### DIFF
--- a/packages/cli/internal/pkg/cli/project_init.go
+++ b/packages/cli/internal/pkg/cli/project_init.go
@@ -125,7 +125,7 @@ func (o *initProjectOpts) validateProject() error {
 func BuildProjectInitCommand() *cobra.Command {
 	vars := initProjectVars{}
 	cmd := &cobra.Command{
-		Use:   "init project_name",
+		Use:   "init project_name --workflow-type my_workflow_type",
 		Short: "Initialize current directory with a new empty AGC project for a particular workflow type.",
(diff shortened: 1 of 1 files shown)


packages/cli/internal/pkg/cli/project_init.go
```diff
@@ -125,7 +125,7 @@ func (o *initProjectOpts) validateProject() error {
 func BuildProjectInitCommand() *cobra.Command {
 	vars := initProjectVars{}
 	cmd := &cobra.Command{
-		Use:   "init project_name",
+		Use:   "init project_name --workflow-type my_workflow_type",
```
tneely, line 128: Workflow type is effectively an enum, so `my_workflow_type` may be misleading to customers since it suggests a freetext field. I'm not sure if this is valid CLI-ese but thoughts on `"init project_name --workflow-type {wdl|nextflow}"`

The Cobra use documentation touches on this a little, but I'm not sure what it considers an argument:

```
//   { } delimits a set of mutually exclusive arguments when one of the arguments is required. If the arguments are
//       optional, they are enclosed in brackets ([ ]).
```
[source](https://pkg.go.dev/github.com/spf13/cobra@v1.2.1#Command.Use)

markjschreiber (approved):
Looks good

tneely (approved):

PR #41, opened 2021-09-29
Remove title from slack payload

tneely:
*Issue #, if available:*

*Description of changes:*
I'm removing the title from the Slack payload. Quotes in the title can escape the workflow payload JSON, causing the Slack workflow to fail. Since Slack expands the PR link with a preview, the title is less important to have in the message.


IllyaYalovyy (approved):
Thank you for fixing it.

PR #44, opened 2021-09-29
Rename println to avoid builtin name clash

tneely:
*Issue #, if available:*

*Description of changes:*
[println](https://pkg.go.dev/builtin#println) is a builtin function in Go. We were overriding that function which could have unintended consequences. I am renaming our function to `printLn` to avoid the conflict.

Additionally, I have replaced lint exception comments with Blank Identifiers to explicitly indicate we don't care about the return values rather than rely on the linter comment.

### DIFF
--- a/packages/cli/internal/pkg/cli/logs_core.go
+++ b/packages/cli/internal/pkg/cli/logs_core.go
@@ -39,7 +39,7 @@ Valid time units are "s", "m", and "h".`
 	tailFlagDescription = "Follow the log output."
 )
 
-var println = fmt.Println
+var printLn = fmt.Println
 
 type logsSharedVars struct {
 	tail        bool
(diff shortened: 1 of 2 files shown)


nbraid (approved):

packages/cli/internal/pkg/cli/logs_core.go
```diff
@@ -39,7 +39,7 @@ Valid time units are "s", "m", and "h".`
 	tailFlagDescription = "Follow the log output."
 )
 
-var println = fmt.Println
+var printLn = fmt.Println
```
IllyaYalovyy, line 42: var printLn = func (args ...intervace{}) {
  _, _ = fmt.Println(args...)
}

This change will make the rest of the code a bit tidier.

IllyaYalovyy (changes requested):

IllyaYalovyy (approved):

PR #50, opened 2021-09-30
Remove hardcoded Node version

a-li:
Remove hardcoded Node version so it can intelligently pick the correct version of Node from `.nvmrc`


elliot-smith (approved):

tneely (approved):
Looks good, thanks!

PR #57, opened 2021-10-01
Ahmaalba/Adapter Role Least Privilege Design

abassyiouni:
**Description**
Removed unnecessary permissions from adapter role, and minimized permissions that it requires for both Cromwell and Nextflow.

**Testing**
All code tested by running `make`
Cromwell code tested by running `bam-to-unmapped-bams`
Nextflow code tested by running `hello`

### DIFF
--- a/packages/cdk/lib/roles/cromwell-adapter-role.ts
+++ b/packages/cdk/lib/roles/cromwell-adapter-role.ts
@@ -4,21 +4,15 @@ import { NextflowDescribeJobsBatchPolicy } from "./policies/nextflow-describe-jo
 import { NextflowSubmitJobBatchPolicy } from "./policies/nextflow-submit-job-batch-policy";
 import { BucketOperations } from "../../common/BucketOperations";
 
-export interface NextflowAdapterRoleProps {
-  headJobDefinitionArn: string;
-  jobQueueArn: string;
+export interface CromwellAdapterRoleProps {
   readOnlyBucketArns: string[];
   readWriteBucketArns: string[];
 }
 
-export class NextflowAdapterRole extends iam.Role {
-  constructor(scope: cdk.Construct, id: string, props: NextflowAdapterRoleProps) {
+export class CromwellAdapterRole extends iam.Role {
+  constructor(scope: cdk.Construct, id: string, props: CromwellAdapterRoleProps) {
     super(scope, id, {
       assumedBy: new iam.ServicePrincipal("ecs-tasks.amazonaws.com"),
-      inlinePolicies: {
-        NextflowDescribeJobsPolicy: new NextflowDescribeJobsBatchPolicy(),
-        NextflowSubmitJobsPolicy: new NextflowSubmitJobBatchPolicy(props),
-      },
     });
 
     BucketOperations.grantBucketAccess(this, this, props.readOnlyBucketArns, true);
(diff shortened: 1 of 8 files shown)


tneely (approved):

packages/cdk/lib/roles/policies/nextflow-submit-job-batch-policy.ts
```diff
@@ -0,0 +1,17 @@
+import * as iam from "monocdk/aws-iam";
+import { NextflowAdapterRoleProps } from "../nextflow-adapter-role";
+
+export class NextflowSubmitJobBatchPolicy extends iam.PolicyDocument {
+  constructor(props: NextflowAdapterRoleProps) {
```
elliot-smith, line 5: NIT: This class should have it's own props and NextflowAdapterRoleProps should extend this classes props instead of the other way around

packages/cdk/lib/roles/cromwell-adapter-role.ts
```diff
@@ -0,0 +1,21 @@
...
+  readWriteBucketArns: string[];
+}
+
+export class CromwellAdapterRole extends iam.Role {
+  constructor(scope: cdk.Construct, id: string, props: CromwellAdapterRoleProps) {
+    super(scope, id, {
+      assumedBy: new iam.ServicePrincipal("ecs-tasks.amazonaws.com"),
+    });
+
+    BucketOperations.grantBucketAccess(this, this, props.readOnlyBucketArns, true);
```
IllyaYalovyy, line 18: Why adapter needs access to S3 buckets? Is it not just pass though requests from a client to an engine?

IllyaYalovyy (changes requested):
I have concerns about granting adapter any access to s3.

packages/cdk/lib/roles/cromwell-adapter-role.ts
wleepang: The Cromwell adapter reads workflow bundles from S3. At minimum, it will need access to a context's S3 location.

IllyaYalovyy: Does it need access to s3 locations defined in data section of the project?

abassyiouni: I tested this by running bam-unmapped-to-mapped with the permissions above and it passed successfully, which uses inputs from other S3 buckets. I am not sure about how the WES Adapter works, but in my experimentation, both nextflow and cromwell adapters need access to the context's s3 location or workflow execution fails.

packages/cdk/lib/roles/policies/nextflow-submit-job-batch-policy.ts
abassyiouni: Adjusted 👍

IllyaYalovyy (approved):

packages/cdk/lib/roles/cromwell-adapter-role.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export interface CromwellAdapterRoleProps {
+  readOnlyBucketArns: string[];
+  readWriteBucketArns: string[];
+}
+
+export class CromwellAdapterRole extends iam.Role {
+  constructor(scope: cdk.Construct, id: string, props: CromwellAdapterRoleProps) {
+    super(scope, id, {
+      assumedBy: new iam.ServicePrincipal("ecs-tasks.amazonaws.com"),
```
nbraid, line 13: Presumably this would be set by default if we just left the taskRole alone.

Any reason to externalize this role entirely rather than do something like `buckets.grantReadAccess(adapter)`

abassyiouni: Just for sake of consistency between the code. Seemed a little odd to have nextflow's role set its own s3 permissions while we use the cromwell-engine-stack to do that for cromwell.

PR #61, opened 2021-10-04
LPD Full Managed Policy Permission Descope

abassyiouni:
***Description***
Removal of all AWS managed policies from AGC cli. Rather, more specific policies, and roles have been implemented to reach a least privilege design.

***Testing***

- All code tested with `make` 
-  Cromwell and nextflow tested with `hello` workflow
- Cromwell tested with workflows `cram-to-bam` and `bam-to-unmapped-bams`
- Nextflow was tested with `rnaseq`
- Cromwell and nextflow were tested with adapter, engine, and workflow log retrieval.
- Cromwell and nextflow were tested by running `workflow stop` while workflows were running.

Integration tests passed locally


packages/cdk/lib/types/engine-options.ts
```diff
@@ -2,6 +2,7 @@ import { RoleProps } from "monocdk/aws-iam";
 import { IJobQueue } from "monocdk/aws-batch";
 import { IVpc } from "monocdk/aws-ec2";
 import { ContextAppParameters } from "../env";
+import {Environment} from "monocdk";
```
tneely, line 5: Please run prettier to format your code (it should pad the barrel import with spaces)

packages/cdk/lib/roles/policies/cromwell-batch-policy.ts
```diff
@@ -0,0 +1,33 @@
...
+export class CromwellBatchPolicy extends iam.PolicyDocument {
+  constructor(props: CromwellBatchPolicyProps) {
+    const cromwellJobArn = `arn:aws:batch:${props.region}:${props.account}:job-definition/cromwell_*`;
+    super({
+      assignSids: true,
+      statements: [
+        new iam.PolicyStatement({
+          effect: iam.Effect.ALLOW,
+          actions: ["batch:DescribeJobDefinitions", "batch:ListJobs", "batch:DescribeJobs", "batch:DescribeJobQueues", "batch:DescribeComputeEnvironments"],
+          resources: ["*"],
```
tneely, line 18: Is it possible to scope these down to a specific job queue where possible? Same for the nextflow batch policy

https://docs.aws.amazon.com/batch/latest/userguide/ExamplePolicies_BATCH.html#iam-example-restrict-job-queue

abassyiouni: Unfortuently not, these are all account level permissions: https://docs.aws.amazon.com/service-authorization/latest/reference/list_awsbatch.html

packages/cdk/lib/types/engine-options.ts
abassyiouni: Will do 👍

tneely (approved):

packages/cdk/lib/roles/policies/nextflow-batch-policy.ts
```diff
@@ -0,0 +1,28 @@
...
+      assignSids: true,
+      statements: [
+        new iam.PolicyStatement({
+          effect: iam.Effect.ALLOW,
+          actions: ["batch:DescribeJobDefinitions", "batch:ListJobs", "batch:DescribeJobs", "batch:DescribeJobQueues", "batch:DescribeComputeEnvironments"],
+          resources: ["*"],
+        }),
+        new iam.PolicyStatement({
+          effect: iam.Effect.ALLOW,
+          actions: ["batch:RegisterJobDefinition"],
```
nbraid, line 22: I think(?) Nextflow will clean up old job definitions when it's done, so we may need DeregisterJobDefinition or whatever as well.

packages/cdk/lib/roles/policies/nextflow-submit-job-batch-policy.ts
```diff
@@ -0,0 +1,24 @@
...
+export class NextflowSubmitJobBatchPolicy extends iam.PolicyDocument {
+  constructor(props: NextflowSubmitJobBatchPolicyProps) {
+    const nextflowJobArn = `arn:aws:batch:${props.region}:${props.account}:job-definition/nf-ubuntu:*`;
+
+    super({
+      assignSids: true,
+      statements: [
+        new iam.PolicyStatement({
+          effect: iam.Effect.ALLOW,
+          actions: ["batch:SubmitJob"],
```
nbraid, line 18: Nextflow will Cancel/Terminate jobs if the parent job is cancelled too. We'll probably want `CancelJob/TerminateJob` permissions.

packages/cdk/lib/stacks/nested/nextflow-engine-stack.ts
```diff
@@ -31,15 +31,15 @@ export class NextflowEngineStack extends NestedEngineStack {
     const outputBucket = Bucket.fromBucketName(this, "OutputBucket", params.outputBucketName);
     const artifactBucket = Bucket.fromBucketName(this, "ArtifactBucket", params.artifactBucketName);
 
-    const adapterRole = new Role(this, "AdapterRole", { assumedBy: new ServicePrincipal("ecs-tasks.amazonaws.com"), ...props.policyOptions });
     const engineRole = new NextflowEngineRole(this, "NextflowEngineRole", {
+      account: props.env?.account ?? "",
```
nbraid, line 35: Does this default work? If we can't figure out our region, then we won't be able to construct an ARN in the policy.

From within a stack you can use `Stack.account` and `Stack.region` to look up these values rather than passing them in via the props. https://docs.aws.amazon.com/cdk/latest/guide/stacks.html

packages/cdk/lib/roles/policies/nextflow-describe-jobs-batch-policy.ts
```diff
@@ -0,0 +1,16 @@
...
+      statements: [
+        new iam.PolicyStatement({
+          effect: iam.Effect.ALLOW,
+          actions: ["batch:DescribeJobs"],
+          resources: ["*"],
+        }),
+      ],
+    });
+  }
+}
```
nbraid, line 16: Does pulling this out into a class clarify much? Just doing this inline seems just as readable and less code to understand.

`adapter.grant('batch:DescribeJobs', '*')` seems just as good.

packages/cdk/lib/roles/nextflow-adapter-role.ts
```diff
@@ -0,0 +1,25 @@
...
+    super(scope, id, {
+      assumedBy: new iam.ServicePrincipal("ecs-tasks.amazonaws.com"),
+      inlinePolicies: {
+        NextflowDescribeJobsPolicy: new NextflowDescribeJobsBatchPolicy(),
+        NextflowSubmitJobsPolicy: new NextflowSubmitJobBatchPolicy(props),
+      },
+    });
+
+    BucketOperations.grantBucketAccess(this, this, props.readOnlyBucketArns, true);
+    BucketOperations.grantBucketAccess(this, this, props.readWriteBucketArns);
```
nbraid, line 23: Why do we need to pass `this` twice?

packages/cdk/lib/roles/cromwell-adapter-role.ts
```diff
@@ -0,0 +1,19 @@
...
+}
+
+export class CromwellAdapterRole extends iam.Role {
+  constructor(scope: cdk.Construct, id: string, props: CromwellAdapterRoleProps) {
+    super(scope, id, {
+      assumedBy: new iam.ServicePrincipal("ecs-tasks.amazonaws.com"),
+    });
+
+    BucketOperations.grantBucketAccess(this, this, props.readOnlyBucketArns, true);
+    BucketOperations.grantBucketAccess(this, this, props.readWriteBucketArns);
```
nbraid, line 17: I assume the `true` is a `isReadOnly` parameter? I kind of prefer CDK's `grantReadWrite` and `grantRead` for clarity.

packages/cdk/lib/roles/policies/nextflow-batch-policy.ts
abassyiouni: From what I've seen, both nextflow and cromwell both don't clean up the job definitions

packages/cdk/lib/roles/policies/nextflow-submit-job-batch-policy.ts
abassyiouni: Will test the workflow stop command and make sure everything works 👍

packages/cdk/lib/stacks/nested/nextflow-engine-stack.ts
abassyiouni: Thanks for pointing that out, I didn't know I can get them from within the stack, will update

packages/cdk/lib/roles/policies/nextflow-describe-jobs-batch-policy.ts
abassyiouni: Sounds good

packages/cdk/lib/roles/cromwell-adapter-role.ts
abassyiouni: Yup its for the readonly parameter, I agree I think it should be more readable. I can push that out in another PR tho since I think it touches a couple other places of code.

packages/cdk/lib/roles/policies/nextflow-describe-jobs-batch-policy.ts
abassyiouni: `adapter.grant()` will actually not work sense you don't define the resource in the grant method, but instead in this case the `adapter` will be the resource. I have pulled this definition out of its own class and made it defined within the role document

packages/cdk/lib/roles/policies/nextflow-submit-job-batch-policy.ts
```diff
@@ -0,0 +1,24 @@
...
+  constructor(props: NextflowSubmitJobBatchPolicyProps) {
+    const nextflowJobArn = `arn:aws:batch:${props.region}:${props.account}:job-definition/nf-*:*`;
+
+    super({
+      assignSids: true,
+      statements: [
+        new iam.PolicyStatement({
+          effect: iam.Effect.ALLOW,
+          actions: ["batch:SubmitJob"],
+          resources: props.submitJobPolicyArns.concat(nextflowJobArn),
```
AndreyDovydenko, line 19: pease prefer language constructs over library functions, like 
```
resources: [...props.submitJobPolicyArns, nextflowJobArn ]
```

packages/cdk/lib/roles/policies/nextflow-submit-job-batch-policy.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export interface NextflowSubmitJobBatchPolicyProps {
+  account: string;
+  region: string;
+  submitJobPolicyArns: string[];
+}
+
+export class NextflowSubmitJobBatchPolicy extends iam.PolicyDocument {
+  constructor(props: NextflowSubmitJobBatchPolicyProps) {
+    const nextflowJobArn = `arn:aws:batch:${props.region}:${props.account}:job-definition/nf-*:*`;
```
AndreyDovydenko, line 11: This is a little concerning that we have to build up the arn in here that way. Is it the best we could do?

packages/cdk/lib/roles/policies/nextflow-submit-job-batch-policy.ts
```diff
@@ -0,0 +1,24 @@
+import * as iam from "monocdk/aws-iam";
+
+export interface NextflowSubmitJobBatchPolicyProps {
+  account: string;
```
AndreyDovydenko, line 4: passing the account and region around is a concerning sign to me, are we expecting those to ever not match the context in which the script runs? E.g. would we see account/region pair change over the runtime?

packages/cdk/lib/roles/policies/nextflow-submit-job-batch-policy.ts
abassyiouni: Will do

packages/cdk/lib/roles/cromwell-engine-role.ts
```diff
@@ -2,20 +2,46 @@ import * as cdk from "monocdk";
...
   readWriteBucketArns: string[];
   policies: PolicyOptions;
+  jobQueueArn: string;
 }
 
 export class CromwellEngineRole extends iam.Role {
   constructor(scope: cdk.Construct, id: string, props: CromwellEngineRoleProps) {
+    const cromwellJobArn = Arn.format(
+      {
+        account: props.account,
```
AndreyDovydenko, line 21: As per documentation at https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_core.Arn.html#static-formatcomponents-stack

```
If partition, region or account are not specified, the stack's partition, region and account will be used.
```

so we don't need to pass `region` / `account` in there, and hence we don't need them to be captured in our props too.

abassyiouni: Sounds good

PR #64, opened 2021-10-04
Display message if no contexts deployed

tneely:
*Issue #, if available:*

*Description of changes:*
The `context status` command prints nothing if no contexts are deployed. I am modifying the command to inform the user when this is the case.

No contexts
```
$ agc context status
2021-10-04T12:50:09-07:00 𝒊  There are no contexts deployed.
```

One context
```
$ agc context status
INSTANCE	myContext		STARTED	true
```

### DIFF
--- a/packages/cli/internal/pkg/cli/context_status.go
+++ b/packages/cli/internal/pkg/cli/context_status.go
@@ -54,7 +54,11 @@ func BuildContextStatusCommand() *cobra.Command {
 			if err != nil {
 				return clierror.New("context status", nil, err)
 			}
-			format.Default.Write(contextInstances)
+			if len(contextInstances) > 0 {
+				format.Default.Write(contextInstances)
+			} else {
+				format.Default.Write("There are no contexts deployed.")
+			}
 			return nil
 		}),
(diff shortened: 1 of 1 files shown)


elliot-smith (approved):

ghawk1ns (approved):

packages/cli/internal/pkg/cli/context_status.go
```diff
@@ -54,7 +55,11 @@ func BuildContextStatusCommand() *cobra.Command {
 			if err != nil {
 				return clierror.New("context status", nil, err)
 			}
-			format.Default.Write(contextInstances)
+			if len(contextInstances) > 0 {
+				format.Default.Write(contextInstances)
```
IllyaYalovyy, line 59: There is a function that is responsible for printing out collections. Will make sense to move this functionality to that function? In this case all commands that output a collection would benefit from this change.

This code could be also simplified:

```go
if len(contextInstances) == 0 {
  log.Info().Msg("There are no contexts deployed.")
}
format.Default.Write(contextInstances)
```

PR #66, opened 2021-10-05
[Bug] Creating a cromwell Spot Context also creates an on demand Compute Env

AndreyDovydenko:
[Bug] Creating a cromwell Spot Context also creates an on demand Compute Env

Separating cromwell and nextflow batch stacks creation sufficiently enough to have separate configuration for the two of them. Creating an on-demand supplement batch env only for nextfow engine.


packages/cdk/lib/stacks/context-stack.ts
```diff
@@ -1,31 +1,129 @@
...
+        createOnDemandBatch: true,
+      };
+    }
+
+    return {
+      ...commonBatchProps,
+      createSpotBatch: false,
+      createOnDemandBatch: true,
+    };
+  }
```
elliot-smith, line 118: Could this not be simplified to;
```
return {
        ...commonBatchProps,
        createSpotBatch: props.contextParameters.requestSpotInstances,
        createOnDemandBatch: true,
      };
```

packages/cdk/lib/stacks/context-stack.ts
```diff
@@ -1,31 +1,129 @@
...
+        createSpotBatch: true,
+        createOnDemandBatch: false,
+      };
+    }
+
+    return {
+      ...commonBatchProps,
+      createSpotBatch: false,
+      createOnDemandBatch: true,
+    };
```
elliot-smith, line 92: Could this not be simplified to;

return {
        ...commonBatchProps,
        createSpotBatch: props.contextParameters.requestSpotInstances,
        createOnDemandBatch: true,
      };

packages/cdk/lib/stacks/nested/batch-stack.ts
```diff
@@ -56,4 +66,15 @@ export class BatchStack extends NestedStack {
       resourceTags: this.nestedStackParent?.tags.tagValues(),
     });
   }
+
+  private getBatchRoles(): IRole[] {
```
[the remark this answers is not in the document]
AndreyDovydenko: I think having a dedicated small function with no side effects is a cleaner way to go, compared to spawning a state capturing instance variable.

packages/cdk/lib/stacks/context-stack.ts
AndreyDovydenko: I think it's a little cleaner to have it this way. I tend to be on "simple is better than smart" part of the spectrum. 

a block like 

```
return {
...commonBatchProps,
createSpotBatch: props.contextParameters.requestSpotInstances,
createOnDemandBatch: !props.contextParameters.requestSpotInstances,,
};
```
could take a second to mentally compile it and figure out what's going on there, where as with the currently proposed code it's pretty simple to read that one case will have only spot and the other would have only on demand.

packages/cdk/lib/stacks/context-stack.ts
AndreyDovydenko: I think here I'd agree that it's an overkill to expand that much

packages/cdk/lib/stacks/context-stack.ts
```diff
@@ -1,31 +1,129 @@
...
+  private renderCromwellStack(props: ContextStackProps) {
+    const batchProps = this.getCromwellBatchProps(props);
+    const batchStack = this.renderBatchStack(batchProps);
+
+    let jobQueue;
+    if (props.contextParameters.requestSpotInstances) {
+      jobQueue = batchStack.batchSpot.jobQueue;
+    } else {
+      jobQueue = batchStack.batchOnDemand.jobQueue;
+    }
```
tneely, line 49: This smells like a leaky abstraction. We've gotten to a state where `BatchStack` is trying to represent two different states based on the compute engine it's using, resulting in complexity across multiple classes to handle the different types of state.

Rather than make `BatchStack` fit every use case, what if we rendered batches on demand?  Then getting out the batch we want for Cromwell would be 

```
const batchStack = this.renderBatchStack(batchProps);
const computeType = props.contextParameters.requestSpotInstances ? ComputeType.SPOT : ComputeType.ON_DEMAND;
const jobQueue = batchStack.renderBatch("TaskBatch", computeType).jobQueue;
```

and Nextflow

```
const batchStack = this.renderBatchStack(batchProps);
let jobQueue, headQueue;
if (props.contextParameters.requestSpotInstances) {
  jobQueue = batchStack.renderBatch("TaskBatch", ComputeType.SPOT).jobQueue;
  headQueue = batchStack.renderBatch("TaskBatch", ComputeType.ON_DEMAND).jobQueue;
} else {
  headQueue = jobQueue = batchStack.renderBatch("TaskBatch", ComputeType.ON_DEMAND).jobQueue;
}
```

packages/cdk/lib/stacks/nested/batch-stack.ts
```diff
@@ -15,35 +16,44 @@ export interface BatchStackProps extends NestedStackProps {
...
+   * Request Spot capacity to be created
+   */
+  readonly createSpotBatch: boolean;
+  /**
+   * Request On-Demand capacity to be created
+   */
+  readonly createOnDemandBatch: boolean;
 }
 
 export class BatchStack extends NestedStack {
```
tneely, line 29: Thoughts on changing this class to render batches through a public method rather than through the constructor? It would simplify a lot of the logic both here and in the context stack above.

```
export class BatchStack extends NestedStack {
  public readonly vpc: IVpc;
  public readonly artifactBucket: IBucket;
  public readonly outputBucket: IBucket;
  public readonly readBucketArns: string[];
  public readonly readWriteBucketArns: string[];

  constructor(scope: Construct, id: string, props: BatchStackProps) {
    super(scope, id, props);

    const { vpc, contextParameters } = props;
    this.vpc = vpc;
    this.artifactBucket = BucketOperations.importBucket(this, "ArtifactBucket", contextParameters.artifactBucketName);
    this.outputBucket = BucketOperations.importBucket(this, "OutputBucket", contextParameters.outputBucketName);

    this.readBucketArns = contextParameters.readBucketArns ?? [];
    this.readWriteBucketArns = contextParameters.readWriteBucketArns ?? [];

    this.instanceTypes = contextParameters.instanceTypes;
  }

  public renderBatch(id: string, computeType?: ComputeType): Batch {
    const batch = new Batch(this, id, {
      vpc: this.vpc,
      instanceTypes: this.instanceTypes,
      computeType,
      launchTemplateData: LAUNCH_TEMPLATE,
      awsPolicyNames: ["AmazonSSMManagedInstanceCore", "CloudWatchAgentServerPolicy"],
      resourceTags: this.nestedStackParent?.tags.tagValues(),
    });

    BucketOperations.grantBucketAccess(this, batch.role, (this.readBucketArns).concat(this.artifactBucket.bucketArn), true);
    BucketOperations.grantBucketAccess(this, batch.role, (this.readWriteBucketArns).concat(this.outputBucket.bucketArn));

    this.artifactBucket.grantRead(batch.role);
    this.outputBucket.grantReadWrite(batch.role);

    return batch;
  }
}
```

AndreyDovydenko: Thanks, i think this is a good idea to not plumb the state through two different stacks. More I look at it less value of a nested stack for batch I see here all together. I'll ping you on that.

AndreyDovydenko: As we have agreed on this class would remain `constructor()` based and consuming it's configuration from props, instead of allowing outside code to create resources within this stack instead by invoking `public renderBatch()` method potentially unbound number of times.

The decision comes from us wanting to keep the nested stacks separated for the time being, the other option discussed has been abolish the nested stack for batch altogether and just render required compute environments where we have our engines rendered.

PR #88, opened 2021-10-12
Tabular text implementation and tests

abassyiouni:
**Description of Changes**

Implemented Tabular format, and modified format.go to set formatter based on `--format` global flag.
Tabular format is compatible with:

- Simple structures with no nested structures
- Structures with embedded slices (slices may include unnested structures, or simple types)

Upon executing a command with the `--format=tabular` flag, the incoming structure is validated to meet the above prerequisites. If the above prerequisites are not met, output is generated with the default `text` writer.

Example:
```
$ agc workflow status
2021-10-12T12:55:31-07:00 𝒊  Showing workflow run(s). Max Runs: 20
WORKFLOWINSTANCE        myContext       34766e89-a862-455e-8e41-aa231e2246e5    true    COMPLETE        2021-10-12T05:39:20Z    hello
WORKFLOWINSTANCE        newContext      106b000b-a5ce-4e57-8bd5-7944b1b7c273    true    COMPLETE        2021-10-12T05:38:38Z    hello
```

```
$ agc workflow status --format=tabular
2021-10-12T12:56:00-07:00 𝒊  Showing workflow run(s). Max Runs: 20
Id                                      WorkflowName    ContextName     State   SubmittedTime           InProject
34766e89-a862-455e-8e41-aa231e2246e5    hello           myContext       COMPLETE2021-10-12T05:39:20Z    true
106b000b-a5ce-4e57-8bd5-7944b1b7c273    hello           newContext      COMPLETE2021-10-12T05:38:38Z    true

```

All code tested with make.
Manually tested `workflow list` `context list` `context deploy` `workflow run` `logs workflow` `context status` and `workflow status` with and without the --format flag

**Checklist**

----


markjschreiber (approved):

packages/cli/internal/pkg/cli/format/format.go
```diff
@@ -1,8 +1,29 @@
...
 
-var Default Formatter = &Text{os.Stdout}
+var Formats = map[string]Formatter{
+	"text":    textWriter,
+	"tabular": tabularWriter,
+}
+var Format string
+
+var textWriter Formatter = &Text{os.Stdout}
+var tabularWriter = &TextTabular{
```
IllyaYalovyy, line 15: We don't have to create all formatters every time we launch AGC.

abassyiouni: Addressed 👍

packages/cli/internal/pkg/cli/format/format.go
```diff
@@ -4,24 +4,28 @@ import (
...
+		return &TextTabular{
+		*tabwriter.NewWriter(os.Stdout, 0, 8, 0, tabularDelimiter[0], 0),
+	}},
 }
-var Default Formatter
 
-func SetDefaultWriter() {
-	if writer, ok := Formats[Format]; ok {
-		Default = writer
+func SetFormatter(format FormatterType) {
```
IllyaYalovyy, line 24: The code looks good. A few final things:
1. If no one calls SetFormatter then Default formatter is nil - not very safe
2. Usage of map here seems to make things a bit more hard to read.

Here is how we can improve it:

```go
var Default = NewText()

// in Text formatter file
func NewText() *Text {
  return &Text{os.Stdout} 
}

// in tabular formatter file
func NewTable () *Table {
  return &Table{tabwriter.NewWriter(os.Stdout, 0, 8, 0, tabularDelimiter[0], 0)}
}

func SetFormatter(format FormatterType) {
  switch format {
  case textFormat: Default = NewText()
  case tableFormatter: Default = NewTable()
  }
}
```

packages/cli/internal/pkg/cli/format/text_tabular.go
```diff
@@ -21,7 +21,8 @@ const (
 func (f *TextTabular) Write(o interface{}) {
```
IllyaYalovyy, line 21: Could you please rename it to just "tabular"? It's a bit confusing whether it text or tabular. Especially taking into account that we already have tabular.

packages/cli/internal/pkg/cli/format/format.go
```diff
@@ -4,24 +4,28 @@ import (
...
 
-var Formats = map[string]Formatter{
-	"text":    textWriter,
-	"tabular": tabularWriter,
-}
-var Format string
+type FormatterType string
+const (
+	textFormat FormatterType = "text"
+	tabularFormat FormatterType = "tabular"
```
IllyaYalovyy, line 12: I think the string value for this should be just "table".

For more details see aws cli output options:
- https://docs.aws.amazon.com/cli/latest/userguide/cli-usage-output-format.html

I would suggest to follow the same names, to avoid confusion.

packages/cli/internal/pkg/cli/format/text_tabular.go
IllyaYalovyy: Or even better to just "table"?

IllyaYalovyy (changes requested):

packages/cli/internal/pkg/cli/format/text_tabular.go
abassyiouni: Adjusted

IllyaYalovyy (approved):

PR #111, opened 2021-10-21
test: Added context deploy benchmarking script

abassyiouni:
**Description of Changes**

Added implementation for context deploy benchmarking script. The script is called with the engine type being tested. 

For example: `./context-benchmark.sh cromwell`. 

If no engine is provided when the script is called, it fails and asks the user to input an engine type. 

For example: 
```
context-deploy-benchmark % ./context-benchmark.sh 
Please specify engine type to benchmark (cromwell | nextflow)
```

Within the script, there is a variable (`CONTEXT_AMOUNT`) defining how many contexts are going to be deployed, and will compute the average time across those contexts. The max value that CONTEXT_AMOUNT can be set to is **5**. Once all contexts are deployed, average deployment time is returned in minutes and seconds.

Example execution:
```
ahmaalba@3c22fb77b124 context-deploy-benchmark % ./context-benchmark.sh cromwell         
2021-10-21T10:15:58-07:00 𝒊  Deploying context(s)
Deploying resources for context 'cromwell1'... [-----------------------------------------------------------------------------------------------------------------------------------------------------------------]17m13s
DETAIL  ...
SUMMARY false   cromwell1
Average cromwell deployment time across 1 contexts is: 17 m 16 s
```

Tested locally and within EC2 t2.micro instances, with single and multime cromwell/nextflow contexts

**Checklist**

----


ghawk1ns:
Should the script also destroy the contexts afterwords?

scripts/context-deploy-benchmark/agc-project.yaml
```diff
@@ -0,0 +1,50 @@
+---
+name: Demo
```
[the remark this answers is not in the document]
abassyiouni: Will do

scripts/context-deploy-benchmark/context-benchmark.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
+
+CONTEXT_AMOUNT=1
+
+run_contexts() {
+  for ((i = 1; i <= CONTEXT_AMOUNT; i++))
```
tneely, line 6: We can deploy the contexts in parallel to reduce bench marking time. AWS load is unpredictable and unknown to us at any given time, so the difference in running parallel vs sequentially will be largely noise

scripts/context-deploy-benchmark/context-benchmark.sh
```diff
@@ -0,0 +1,31 @@
...
+  do
+    agc context deploy -c "$1$i";
```
tneely, line 8: Capturing single context deployments is important, but we should also try to capture AGC's multi-context performance in a separate benchmark (e.g. `agc context deploy --all` or `agc context deploy -c "context1" -c "context2" ...`).

Today this almost equivalent to running them sequentially, but we have future enhancements coming that will be able to deploy them in parallel.

ghawk1ns (approved):

PR #129, opened 2021-10-29
fix: Adds a message when new logs aren't shown to the user immediately

elliot-smith:
Issue #, if available: NA

**Description of Changes**

Currently, when you run the tail logs command it can look like the command hung when it is run. This change provides some messaging to the user so that they are aware that the reason that there are no logs shows is because there aren't any new ones.

NB: The current approach *can* show the message immediately before messages are shown if they exist since we split up the request to AWS if the number of streams is larger than the limit. However, this is unlikely, could also happen if the logs start appearing immediately after the first request is made and would make the logic more complicated because of this I have gone with the simpler solution.

Ran the commands while running workflows

```
$ agc logs engine -c onDemandContext -t
2021-10-29T10:53:41-07:00 𝒊  Showing engine logs for 'onDemandContext'
There are no new logs. Please wait for new logs to show...
```

**Checklist**

- [ ] I have linted my code before raising the PR
- [ ] Title of this Pull Request follows Conventional Commits standard: https://www.conventionalcommits.org/en/v1.0.0/

----

### DIFF
--- a/packages/cli/internal/pkg/cli/logs_core.go
+++ b/packages/cli/internal/pkg/cli/logs_core.go
@@ -124,6 +124,19 @@ func (o *logsSharedOpts) followLogGroup(logGroupName string) error {
 }
 
 func (o *logsSharedOpts) displayEventFromChannel(channel <-chan cwl.StreamEvent) error {
+	firstEvent := <-channel
+
+	if firstEvent.Err != nil {
+		return firstEvent.Err
+	}
+	if len(firstEvent.Logs) > 0 {
+		for _, line := range firstEvent.Logs {
+			printLn(line)
+		}
+	} else {
+		printLn("There are no new logs. Please wait for new logs to show...")
+	}
+
 	for event := range channel {
 		if event.Err != nil {
 			return event.Err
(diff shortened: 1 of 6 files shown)


packages/cli/internal/pkg/cli/logs_core.go
```diff
@@ -124,6 +124,19 @@ func (o *logsSharedOpts) followLogGroup(logGroupName string) error {
...
+		return firstEvent.Err
+	}
+	if len(firstEvent.Logs) > 0 {
+		for _, line := range firstEvent.Logs {
+			printLn(line)
+		}
+	} else {
+		printLn("There are no new logs. Please wait for new logs to show...")
+	}
+
```
ghawk1ns, line 139: This replicates the existing logic below, we only need to conditionally show a message in the else block of the first event, can we simplify that with a flag?

```
var firstEvent = true
....
   if (firstEvent) {
      printLn("There are no new logs. Please wait for new logs to show...")
   } else {
      log.Debug().Msg("No new logs")
   }
...
firstEvent = false
```

PR #131, opened 2021-11-01
fix: Adds a message when new logs aren't shown to the user immediately

elliot-smith:
Issue #, if available: NA

Description of Changes

Currently, when you run the tail logs command it can look like the command hung when it is run. This change provides some messaging to the user so that they are aware that the reason that there are no logs shows is because there aren't any new ones. This also adds back the logic to send through empty logs when no new events are sent back from Cloudwatch

NB: The current approach can show the message immediately before messages are shown if they exist since we split up the request to AWS if the number of streams is larger than the limit (greater than 100 streams). However this is unlikely and could also happen if the logs start appearing immediately after the first request is made to AWS. Since this could occur naturally anyway, would make the logic more complicated and doesn't impact the customer experience I have gone with the simpler solution.

Description of how you validated changes
Ran the commands while running workflows

$ agc logs engine -c onDemandContext -t
2021-10-29T10:53:41-07:00 𝒊  Showing engine logs for 'onDemandContext'
There are no new logs. Please wait for the first logs to appear...

**Checklist**

----

### DIFF
--- a/packages/cli/internal/pkg/aws/cwl/stream_logs.go
+++ b/packages/cli/internal/pkg/aws/cwl/stream_logs.go
@@ -38,6 +38,8 @@ func (c Client) StreamLogs(ctx context.Context, logGroupName string, streams ...
 			if aws.ToString(lastToken) != aws.ToString(output.NextToken) {
 				stream <- StreamEvent{Logs: parseEventLogs(output.Events, startTime)}
 				lastToken = output.NextToken
+			} else {
+				stream <- StreamEvent{}
 			}
 
 			select {
--- a/packages/cli/internal/pkg/cli/logs_core.go
+++ b/packages/cli/internal/pkg/cli/logs_core.go
@@ -42,6 +42,8 @@ Use a question mark for OR, such as "?ERROR ?WARN". Filter out terms with a minu
 	tailFlagDescription = "Follow the log output."
 )
 
+var logInfo = log.Info
+
 var printLn = func(args ...interface{}) {
 	_, _ = fmt.Println(args...)
 }
@@ -137,7 +139,7 @@ func (o *logsSharedOpts) displayEventFromChannel(channel <-chan cwl.StreamEvent)
 			firstEvent = false
 		} else if firstEvent {
 			firstEvent = false
-			printLn("There are no new logs. Please wait for the first logs to appear...")
+			logInfo().Msg("There are no new logs. Please wait for the first logs to appear...")
 		} else {
 			log.Debug().Msg("No new logs")
 		}
--- a/packages/cli/internal/pkg/cli/logs_core.go
+++ b/packages/cli/internal/pkg/cli/logs_core.go
@@ -136,13 +136,13 @@ func (o *logsSharedOpts) displayEventFromChannel(channel <-chan cwl.StreamEvent)
 			for _, line := range event.Logs {
 				printLn(line)
 			}
-			firstEvent = false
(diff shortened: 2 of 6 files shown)


elliot-smith:
In regards to the AWS Codebuild comment above. It currently isn't able to find the Go version 17.2 so should be ignored for now as it does not relate to anything in the current CR.

```
go-build: definition not found: 1.17.2

See all available versions with 'goenv install --list'.

If the version you need is missing, try upgrading goenv:

  cd /root/.goenv/plugins/go-build/../.. && git pull && cd -

[Container] 2021/11/01 18:44:36 Command did not exit successfully goenv install exit status 2
[Container] 2021/11/01 18:44:36 Phase complete: INSTALL State: FAILED
[Container] 2021/11/01 18:44:36 Phase context status code: COMMAND_EXECUTION_ERROR Message: Error while executing command: goenv install. Reason: exit status 2

```

packages/cli/internal/pkg/cli/logs_core.go
```diff
@@ -132,10 +134,15 @@ func (o *logsSharedOpts) displayEventFromChannel(channel <-chan cwl.StreamEvent)
 			for _, line := range event.Logs {
 				printLn(line)
 			}
+			firstEvent = false
+		} else if firstEvent {
+			firstEvent = false
+			printLn("There are no new logs. Please wait for the first logs to appear...")
```
tneely, line 140: Can this be an info level log? Since it goes to stderr users can capture just the cwl logs from stdout

tneely (changes requested):

packages/cli/internal/pkg/aws/cwl/stream_logs.go
```diff
@@ -38,6 +38,8 @@ func (c Client) StreamLogs(ctx context.Context, logGroupName string, streams ...
 			if aws.ToString(lastToken) != aws.ToString(output.NextToken) {
 				stream <- StreamEvent{Logs: parseEventLogs(output.Events, startTime)}
 				lastToken = output.NextToken
+			} else {
```
IllyaYalovyy, line 41: Could you please explain this change?

packages/cli/internal/pkg/cli/logs_core.go
```diff
@@ -132,10 +136,15 @@ func (o *logsSharedOpts) displayEventFromChannel(channel <-chan cwl.StreamEvent)
 			for _, line := range event.Logs {
 				printLn(line)
 			}
+			firstEvent = false
+		} else if firstEvent {
+			firstEvent = false
```
IllyaYalovyy, line 141: Can this code duplication be avoid?

tneely (approved):

packages/cli/internal/pkg/aws/cwl/stream_logs.go
elliot-smith: Currently, we don't show any messages whenever an empty log comes through which was a regression from a previous change. This is particularly evident when you look at the debug statement in the logs_core package as the statement never triggered.

packages/cli/internal/pkg/cli/logs_core.go
elliot-smith: We could set the value once for every log event but that is less efficient since it would trigger every second. I can change it if required but omitted it consciously.

IllyaYalovyy (approved):

tneely (approved):

PR #138, opened 2021-11-02
fix: Workflow status now ignores unqueryable stacks

elliot-smith:
Issue #, if available: NA

**Description of Changes**

Whenever a customer runs `agc workflow status` while a stack is actively being destroyed or has been destroyed, AGC will still try to query WES to determine the status of workflows that have been ran. This change removes stacks that are currently being removed so that it follows the same logic as stacks that have been removed since we can't trust the status of those stacks.

1. Deployed two contexts
2. Ran two workflows on each of the contexts
3. Waited for the workflows to finish
4. Destroyed one of the contexts and while it was being destroyed, ran `agc workflow status`

```
$ agc workflow status
2021-11-02T09:24:16-07:00 𝒊  Showing workflow run(s). Max Runs: 20
WORKFLOWINSTANCE	spotContext	05a40f99-9f88-4ebd-ab73-730dd4cb3861	true	COMPLETE	2021-11-02T03:57:33Z	hello
```

**Checklist**

----

### DIFF
--- a/packages/cli/internal/pkg/aws/cfn/stack_filters.go
+++ b/packages/cli/internal/pkg/aws/cfn/stack_filters.go
@@ -2,41 +2,44 @@ package cfn
 
 import "github.com/aws/aws-sdk-go-v2/service/cloudformation/types"
 
-var ActiveStacksFilter = []types.StackStatus{
-	types.StackStatusCreateInProgress,
-	types.StackStatusCreateComplete,
-	types.StackStatusRollbackInProgress,
-	types.StackStatusRollbackFailed,
-	types.StackStatusRollbackComplete,
-	types.StackStatusDeleteInProgress,
-	types.StackStatusDeleteFailed,
-	types.StackStatusUpdateInProgress,
-	types.StackStatusUpdateCompleteCleanupInProgress,
-	types.StackStatusUpdateComplete,
-	types.StackStatusUpdateRollbackInProgress,
-	types.StackStatusUpdateRollbackFailed,
-	types.StackStatusUpdateRollbackCompleteCleanupInProgress,
-	types.StackStatusUpdateRollbackComplete,
-	types.StackStatusReviewInProgress,
-	types.StackStatusImportInProgress,
-	types.StackStatusImportComplete,
-	types.StackStatusImportRollbackInProgress,
-	types.StackStatusImportRollbackFailed,
-	types.StackStatusImportRollbackComplete,
+type StackOptions struct {
+	activeStack    bool
+	queryableStack bool
 }
-var QueryableStacksMap = map[types.StackStatus]bool{
-	types.StackStatusCreateComplete:                          true,
-	types.StackStatusUpdateInProgress:                        true,
-	types.StackStatusUpdateCompleteCleanupInProgress:         true,
-	types.StackStatusUpdateComplete:                          true,
-	types.StackStatusUpdateRollbackInProgress:                true,
-	types.StackStatusUpdateRollbackFailed:                    true,
-	types.StackStatusUpdateRollbackCompleteCleanupInProgress: true,
-	types.StackStatusUpdateRollbackComplete:                  true,
-	types.StackStatusReviewInProgress:                        true,
-	types.StackStatusImportInProgress:                        true,
-	types.StackStatusImportComplete:                          true,
-	types.StackStatusImportRollbackInProgress:                true,
-	types.StackStatusImportRollbackFailed:                    true,
-	types.StackStatusImportRollbackComplete:                  true,
+
+var stackDefinitions = map[types.StackStatus]StackOptions{
+	types.StackStatusCreateInProgress:                        {activeStack: true, queryableStack: false},
+	types.StackStatusCreateComplete:                          {activeStack: true, queryableStack: true},
+	types.StackStatusRollbackInProgress:                      {activeStack: true, queryableStack: false},
+	types.StackStatusRollbackFailed:                          {activeStack: true, queryableStack: false},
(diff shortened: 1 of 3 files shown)


tneely (approved):

packages/cli/internal/pkg/aws/cfn/stack_filters.go
```diff
@@ -24,3 +24,20 @@ var ActiveStacksFilter = []types.StackStatus{
...
+	types.StackStatusUpdateRollbackFailed:                    true,
+	types.StackStatusUpdateRollbackCompleteCleanupInProgress: true,
+	types.StackStatusUpdateRollbackComplete:                  true,
+	types.StackStatusReviewInProgress:                        true,
+	types.StackStatusImportInProgress:                        true,
+	types.StackStatusImportComplete:                          true,
+	types.StackStatusImportRollbackInProgress:                true,
+	types.StackStatusImportRollbackFailed:                    true,
+	types.StackStatusImportRollbackComplete:                  true,
+}
```
nbraid, line 42: Could this be merged with the above list?

```go
type StackStatusInfo struct {
	active    bool
	queryable bool
}

var StackStatusInfos = map[string]StackStatusInfo{
	"ACTIVE": {active: true, queryable: true},
}
```

packages/cli/internal/pkg/cli/workflow/workflow_status_test.go
```diff
@@ -395,6 +395,52 @@ func (s *WorkflowStatusTestSuite) TestStatusWorkflow_NonActiveContext() {
 	}
 }
 
+func (s *WorkflowStatusTestSuite) TestStatusWorkflow_NonActiveContexts() {
```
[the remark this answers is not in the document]
elliot-smith: I can see some ways to merge it but the file will still probably be as long as the lengthy sections are the data we are expecting/returning which are largely different.

packages/cli/internal/pkg/aws/cfn/stack_filters.go
elliot-smith: I actually think that the list above is incorrect (deleting/deploying stacks are considered "active" as well as stacks that failed to delete which could have any number of resources inside of them) but wanted to have a discussion about it as a group.

packages/cli/internal/pkg/cli/workflow/workflow_status_test.go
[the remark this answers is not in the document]
elliot-smith: Resolved some of the conflicting changes but the number of mocks differ between the tests and I've experienced issues before when we had custom mocking extracted in helper functions mainly due to extensibility

packages/cli/internal/pkg/aws/cfn/stack_filters.go
```diff
@@ -2,25 +2,44 @@ package cfn
...
+var ActiveStacksFilter []types.StackStatus
+var QueryableStacksMap map[types.StackStatus]bool
+
+func init() {
+	QueryableStacksMap = make(map[types.StackStatus]bool)
+	for stackStatus, stackOptions := range stackDefinitions {
+		if stackOptions.queryableStack {
+			QueryableStacksMap[stackStatus] = true
+		}
```
ghawk1ns, line 40: Is this the same as 

```
QueryableStacksMap[stackStatus] = stackOptions.queryableStack
```

packages/cli/internal/pkg/cli/workflow/manager.go
```diff
@@ -331,15 +331,17 @@ func (m *Manager) isContextDeployed(contextName string) bool {
...
 	if err != nil {
 		if errors.Is(err, cfn.StackDoesNotExistError) {
 			return false
 		}
 		m.err = err
 		return false
 	}
-	return true
+
+	_, activeStatusFlag := cfn.QueryableStacksMap[status]
```
ghawk1ns, line 343: Can you get the same info from 

```
cfn.stackDefinitions[status].activeStack
```

elliot-smith: We can't because the current definition of "active" is not actually "active". I am going to be raising this on Monday with some more succinct points

packages/cli/internal/pkg/aws/cfn/stack_filters.go
```diff
@@ -2,25 +2,42 @@ package cfn
...
+	types.StackStatusImportInProgress:                        {activeStack: true, queryableStack: true},
+	types.StackStatusImportComplete:                          {activeStack: true, queryableStack: true},
+	types.StackStatusImportRollbackInProgress:                {activeStack: true, queryableStack: true},
+	types.StackStatusImportRollbackFailed:                    {activeStack: true, queryableStack: true},
+	types.StackStatusImportRollbackComplete:                  {activeStack: true, queryableStack: true},
+}
+var ActiveStacksFilter []types.StackStatus
+var QueryableStacksMap map[types.StackStatus]bool
+
+func init() {
```
tneely, line 35: I'm not seeing anything ever calling init here. Is there some Go magic that calls it when the package is used?

elliot-smith: It's some go magic :)
Docs: https://tutorialedge.net/golang/the-go-init-function/

tneely (approved):

packages/cli/internal/pkg/aws/cfn/stack_filters.go
IllyaYalovyy: `Init()` is a special function that is called one upon loading that module.

IllyaYalovyy (approved):

PR #139, opened 2021-11-02
perf: Allows customers to deploy and destroy contexts in parallel

elliot-smith:
**Description of Changes**
This change allows us to deploy and destroy contexts in parallel by utilizing different folders as output folders within CDK (a previous commit). The change also updates the process bar since it now needs to be aware of multiple deployments occurring at the same time. We also now print out the entire consecutive log for all errors so that it is easier for users to determine what went wrong during the deployment.

Deployed several contexts and destroyed several contexts. During the deployment phase I also tore down some contexts so that they would fail and the bar re-adjusted it's size and showed the relevant progress. 

```
$ agc context deploy --all
2021-11-02T13:13:13-07:00 𝒊  Deploying context(s)
Deploying resources for context(s) [onDemandContext spotContext] [-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------]20m52s
DETAIL	Agc-Context-NextflowDemo-elliotsm3WJrU7-onD-nextflowNestedStacknextflowNestedStackReso-UH8D4UGKOPYD-ApiProxyAccessLogGroup0E050698-mDQaPXosXLMk	s3://agc-383881756834-us-west-2/project/NextflowDemo/userid/elliotsm3WJrU7/context/onDemandContext	Agc-Context-NextflowDemo-elliotsm3WJrU7-onD-nextflowNestedStacknextflowNestedStackReso-UH8D4UGKOPYD-NextflowEngineEngineLogGroup28B14E87-kCCQpAReA8lg	STARTED		Agc-Context-NextflowDemo-elliotsm3WJrU7-onD-nextflowNestedStacknextflowNestedStackReso-UH8D4UGKOPYD-AdapterLogGroup17C8D427-BjYBSDFu8AUX	https://bkr7q7pow6.execute-api.us-west-2.amazonaws.com/prod/
SUMMARY	false	256	onDemandContext
DETAIL	Agc-Context-NextflowDemo-elliotsm3WJrU7-spo-nextflowNestedStacknextflowNestedStackReso-OSE1HDG58RIE-ApiProxyAccessLogGroup0E050698-16JQnVa2137D	s3://agc-383881756834-us-west-2/project/NextflowDemo/userid/elliotsm3WJrU7/context/spotContext	Agc-Context-NextflowDemo-elliotsm3WJrU7-spo-nextflowNestedStacknextflowNestedStackReso-OSE1HDG58RIE-NextflowEngineEngineLogGroup28B14E87-93Msw7QTMIIG	STARTED		Agc-Context-NextflowDemo-elliotsm3WJrU7-spo-nextflowNestedStacknextflowNestedStackReso-OSE1HDG58RIE-AdapterLogGroup17C8D427-lX73Y7P7mwur	https://whv0rxp7w3.execute-api.us-west-2.amazonaws.com/prod/
SUMMARY	true	256	spotContext
$ agc context destroy --all
2021-11-02T13:45:36-07:00 𝒊  Destroying context(s)'
Destroying resources for context(s) [onDemandContext spotContext] [__________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________________]10m39s
$ agc
```

**Checklist**

----


packages/cli/internal/pkg/aws/cdk/progress.go
```diff
@@ -41,19 +49,139 @@ func (p ProgressStream) DisplayProgress(description string) error {
...
+func combineProgressEvents(progressEventChannels []ProgressStream) (ProgressStream, []*Result) {
+	var results []*Result
+	receiver := make(ProgressStream)
+	var waitGroup sync.WaitGroup
+	waitGroup.Add(len(progressEventChannels))
+
+	for _, channel := range progressEventChannels {
+		progressResult := &Result{}
+		results = append(results, progressResult)
+		go sendDataToReceiverAndUpdateResult(channel, progressResult, &waitGroup, receiver)
```
tneely, line 117: Feels like we have too many layers of interdependent go routines running here. Could we think of a way to simplify here? Perhaps by redefining the data structures around ProgressStream to allow for easier pooling of events across deployments?

packages/cli/internal/pkg/cli/context/manager.go
```diff
@@ -218,3 +226,25 @@ func (m *Manager) readConfig() {
...
+
+func (m *Manager) readProjectInformation() {
+	if m.err != nil {
+		return
+	}
+	m.readProjectSpec()
+	m.readConfig()
+}
+
+func (m *Manager) executeCdkHelper(allProgressStreams []cdk.ProgressStream, description string) {
```
tneely, line 238: perhaps rename to `processExecution` or something more meaningful here

packages/cli/internal/pkg/cli/context/util.go
```diff
@@ -9,3 +13,17 @@ func environmentMapToList(environmentMap map[string]string) []string {
 	}
 	return environmentList
 }
+
+func cdkResultToContextResult(cdkResults []cdk.Result) []ProgressResult {
```
tneely, line 17: nit: `cdkResultToContextResult` -> `cdkResultsToContextResults`

packages/cli/internal/pkg/cli/context/manager_deploy.go
```diff
@@ -5,23 +5,63 @@ import (
...
+func (m *Manager) getStreamsForCdkDeployments(contexts []string) ([]cdk.ProgressStream, []string) {
+	var progressStreams []cdk.ProgressStream
+	var contextsWithStreams []string
+	for _, contextName := range contexts {
+		m.readContextSpec(contextName)
+		m.setCdkConfigurationForDeployment(contextName)
+		m.clearCdkContext(contextDir)
+		m.setContextEnv(contextName)
+
+		if m.err == nil {
```
tneely, line 42: You can move this into `m.deployContext` to reduce nesting - just return immediately `if m.err != nil`

packages/cli/internal/pkg/cli/context/manager_deploy.go
```diff
@@ -5,23 +5,63 @@ import (
...
+			progressStream := m.deployContext(contextName)
+			if progressStream != nil {
+				progressStreams = append(progressStreams, progressStream)
+				contextsWithStreams = append(contextsWithStreams, contextName)
+			}
+		}
+
+		if m.err != nil {
+			m.progressResults = append(m.progressResults, ProgressResult{Context: contextName, Err: m.err})
+		}
```
tneely, line 52: Can we move this into `deployContext`? It's already setting `m.progressResults` on failure so it would be within its scope to also set it when `m.err != nil`

packages/cli/internal/pkg/cli/context/manager_destroy.go
```diff
@@ -11,13 +11,36 @@ const (
...
+		m.setContextPlaceholders()
+		if m.err == nil {
+			progressStream := m.destroyContext(contextName)
+			if progressStream != nil {
+				progressStreams = append(progressStreams, progressStream)
+				contextsWithStreams = append(contextsWithStreams, contextName)
+			}
+		}
+		if m.err != nil {
+			m.progressResults = append(m.progressResults, ProgressResult{Context: contextName, Err: m.err})
```
tneely, line 38: Similar comments here as in deploy - move error handling into `destroyContexts`

packages/cli/internal/pkg/cli/context_deploy.go
```diff
@@ -89,45 +84,94 @@ func (o *deployContextOpts) validateSuppliedContexts(contextList []string) error
...
+	if err != nil {
+		return nil, err
+	}
+
+	sortContextDetails(contextDetails)
+	return contextDetails, nil
+}
+
+func (o *deployContextOpts) deployContexts() error {
+	progressResults := o.ctxManagerFactory().Deploy(o.contexts)
```
tneely, line 103: I don't believe we need the factory any more now that you're handling parallelization inside the `Deploy` function

packages/cli/internal/pkg/cli/context_deploy.go
```diff
@@ -89,45 +84,94 @@ func (o *deployContextOpts) validateSuppliedContexts(contextList []string) error
...
 	aggregateSuggestions := make([]string, 0)
-	for i, result := range results {
-		if result.err != nil {
+
+	failedDeployments := make([]context.ProgressResult, 0)
```
tneely, line 106: length isn't needed here as a second arg in `make` if it's just going to be zero - go will use sensible defaults

packages/cli/internal/pkg/cli/context_deploy.go
```diff
@@ -89,45 +84,94 @@ func (o *deployContextOpts) validateSuppliedContexts(contextList []string) error
...
+		}
+	}
+
+	failedDeploymentsLength := len(failedDeployments)
+	if failedDeploymentsLength > 0 {
+		for i, failedDeployment := range failedDeployments {
+			log.Error().Msgf("Failed to deploy context '%s'. Below is the log for that deployment", failedDeployment.Context)
+
+			outputsLength := len(failedDeployment.Outputs)
+			for i := 0; i < outputsLength-1; i++ {
```
tneely, line 119: Careful with reusing `i` here. It's technically in its own scope and safe, but dangerous and confusing to developers who are trying to access the other `i`

packages/cli/internal/pkg/cli/context_deploy.go
```diff
@@ -89,45 +84,94 @@ func (o *deployContextOpts) validateSuppliedContexts(contextList []string) error
...
+				log.Error().Msg(failedDeployment.Outputs[i])
```
tneely, line 120: If 3/3 deployments fail are we going to dump all the CDK logs into the terminal? That seems unmanageable to users. Half of it would probably get cut off due to terminal window limits. Perhaps if `failedDeploymentsLength > 1` we direct users to CloudFormation?

packages/cli/internal/pkg/cli/context_deploy.go
```diff
@@ -89,45 +84,94 @@ func (o *deployContextOpts) validateSuppliedContexts(contextList []string) error
...
+			if i < failedDeploymentsLength-1 {
+				log.Error().Msgf("%s \n\n\n", failedDeployment.Outputs[outputsLength-1])
+			} else {
+				log.Error().Msg(failedDeployment.Outputs[outputsLength-1])
+			}
+
 			var actionableError *actionableerror.Error
-			ok := errors.As(result.err, &actionableError)
+			ok := errors.As(failedDeployment.Err, &actionableError)
 			if ok {
```
tneely, line 131: You can bring this into the if statement itself:

```go
if errors.As(failedDeployment.Err, &actionableError) {
  // ...
}
```

packages/cli/internal/pkg/cli/context_deploy.go
```diff
@@ -89,45 +84,94 @@ func (o *deployContextOpts) validateSuppliedContexts(contextList []string) error
...
-	results := make([]deployResult, len(contexts))
-	for i, contextName := range contexts {
-		log.Debug().Msgf("Deploying context '%s'", contextName)
-		// TODO: Run in parallel once CDK resolves race condition causing context bleed
-		//       https://github.com/aws/aws-cdk/issues/14350
-		func(ctxManager context.Interface, i int, contextName string) {
-			_ = ctxManager.Deploy(contextName, true)
+func (o *deployContextOpts) validateDeploymentResults() ([]context.Detail, error) {
+	contextDetails, deploymentHasErrors := make([]context.Detail, len(o.contexts)), false
+	aggregateSuggestions := make([]string, 0)
```
tneely, line 145: 0 length isn't needed

packages/cli/internal/pkg/cli/context_deploy.go
```diff
@@ -89,45 +84,94 @@ func (o *deployContextOpts) validateSuppliedContexts(contextList []string) error
...
 
-func (o *deployContextOpts) deployContexts(contexts []string) []deployResult {
-	results := make([]deployResult, len(contexts))
-	for i, contextName := range contexts {
-		log.Debug().Msgf("Deploying context '%s'", contextName)
-		// TODO: Run in parallel once CDK resolves race condition causing context bleed
-		//       https://github.com/aws/aws-cdk/issues/14350
-		func(ctxManager context.Interface, i int, contextName string) {
-			_ = ctxManager.Deploy(contextName, true)
+func (o *deployContextOpts) validateDeploymentResults() ([]context.Detail, error) {
```
tneely, line 143: Is this needed? I can't think of a case where CDK would succeed but Cfn fails

packages/cli/internal/pkg/cli/context_deploy.go
```diff
@@ -89,45 +84,94 @@ func (o *deployContextOpts) validateSuppliedContexts(contextList []string) error
...
+
+	for i, contextName := range o.contexts {
+		go func(ctxManager context.Interface, i int, contextName string) {
 			info, err := ctxManager.Info(contextName)
-			results[i] = deployResult{contextName: contextName, info: info, err: err}
+			contextDetails[i] = info
+			if err != nil {
+				var actionableError *actionableerror.Error
+				ok := errors.As(err, &actionableError)
+				if ok {
```
tneely, line 156: this can be simplified to `if errors.As...`

packages/cli/internal/pkg/cli/context_deploy.go
```diff
@@ -89,45 +84,94 @@ func (o *deployContextOpts) validateSuppliedContexts(contextList []string) error
...
+					log.Error().Err(actionableError.Cause).Msgf(actionableError.Error())
+					aggregateSuggestions = append(aggregateSuggestions, actionableError.SuggestedAction)
```
tneely, line 158: I'm trying to think of when an actionable error from Info would ever be useful here for Deploy but I'm failing to come up with any

tneely (changes requested):

packages/cli/internal/pkg/aws/cdk/progress.go
elliot-smith: I looked through this last night and struggled to simplify the pattern here. The problem is that we are executing several commands in different shells that we are reading which require different event streams. To follow this via a single UI mechanism (such as the receiver), we need to combine the different streams together so that the messaging is concise (this is also required for the Progress Bar). As a result, we need these several Go routines. 

It's also why this previously approved PR required them as well: https://github.com/aws/amazon-genomics-cli/pull/114/files
```
	waitGroup.Add(len(channels))
	for _, c := range channels {
		go multiplexFunc(c)
	}

	go func() {
		waitGroup.Wait()
		close(multiplexedChannel)
	}()
```

packages/cli/internal/pkg/aws/cdk/progress.go
```diff
@@ -41,19 +49,139 @@ func (p ProgressStream) DisplayProgress(description string) error {
...
+
+	var results []Result
+	for _, progressResult := range progressResults {
+		results = append(results, *progressResult)
+	}
+
+	return results
+}
+
+func combineProgressEvents(progressEventChannels []ProgressStream) (ProgressStream, []*Result) {
```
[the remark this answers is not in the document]
elliot-smith: I've refactored the base implementation a little bit and have also modified the CDK code as well to resolve this comment (will leave it open to let you verify whether it is addressed or not)

packages/cli/internal/pkg/aws/cdk/execute_cdk.go
```diff
@@ -74,11 +74,13 @@ func deleteCDKOutputDir(cdkOutputDir string) {
...
 }
 
-func processOutputs(stdout *bufio.Scanner, stderr *bufio.Scanner) (chan ProgressEvent, *sync.WaitGroup) {
+func processOutputs(stdout *bufio.Scanner, stderr *bufio.Scanner, uniqueKey string) (chan ProgressEvent, *sync.WaitGroup) {
 	var wait sync.WaitGroup
 	wait.Add(2)
 	progressChan := make(chan ProgressEvent)
-	currentEvent := &ProgressEvent{}
+	currentEvent := &ProgressEvent{
+		UniqueKey: uniqueKey,
```
tneely, line 82: Rather than piping the unique key down to here, could we generate it here or just one level above? We're basically trying to tie a single CDK execution to a specific event channel so that we can differentiate between each while displaying the progress bar. Unless we need to specifically look up the channel by its key later a simple `rand.Int()` should suffice

packages/cli/internal/pkg/cli/context/manager_deploy.go
```diff
@@ -31,36 +66,17 @@ func (m *Manager) clearCdkContext(appDir string) {
...
 	if m.err != nil {
-		return
+		m.progressResults = append(m.progressResults, ProgressResult{Context: contextName, Err: m.err})
+		return nil
 	}
-	progressStream, err := cmd()
-	if err != nil {
-		m.err = err
-		return
+	contextCmd := func() (cdk.ProgressStream, error) {
```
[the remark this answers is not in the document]
elliot-smith: I think it is as well but it was also the existing logic (see line 46). I'll remove it.

packages/cli/internal/pkg/cli/context_deploy.go
elliot-smith: I've experienced the invalid credential when deploying contexts after they succeeded. It's also safer incase we do actually pass through an actionable error.

packages/cli/internal/pkg/aws/cdk/execute_cdk.go
elliot-smith: We don't just use it for that value, we also use it to tie the outputs together (since they could be any one of them). If you look at the function "combineProgressEvents", we create an empty object called "progressResult" which we add to a results list. This results value is then updated by the ProgressEvent stream which requires the uniqueKey value.

tneely (approved):
Looks good! I think we need to take a look at the context manager at some point (it's getting out of control), but that problem is separate from these changes

packages/cli/internal/pkg/aws/cdk/progress.go
```diff
@@ -41,19 +50,145 @@ func (p ProgressStream) DisplayProgress(description string) error {
...
+	return results
+}
+
+func updateResultFromStream(stream ProgressStream, progressResult *Result, wait *sync.WaitGroup) {
+	defer wait.Done()
+	var lastEvent ProgressEvent
+
+	for event := range stream {
+		if event.Err != nil {
+			progressResult.Err = event.Err
```
IllyaYalovyy, line 80: Are we overwriting events data over and over again? (e.g. error will be overwritten and effectively lost.) What is the purpose of this function?

packages/cli/internal/pkg/aws/cdk/progress.go
```diff
@@ -41,19 +50,145 @@ func (p ProgressStream) DisplayProgress(description string) error {
...
+		}
+	}
+
+	progressResult.Outputs = lastEvent.Outputs
+	progressResult.UniqueKey = lastEvent.UniqueKey
+}
+
+func DisplayProgressBar(description string, progressEvents []ProgressStream) []Result {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
```
IllyaYalovyy, line 92: Why exactly we need cancellation?

packages/cli/internal/pkg/aws/cdk/progress.go
```diff
@@ -41,19 +50,145 @@ func (p ProgressStream) DisplayProgress(description string) error {
...
+
+func combineProgressEvents(progressEventChannels []ProgressStream) (ProgressStream, []*Result) {
+	var results []*Result
+	receiver := make(ProgressStream)
+	var waitGroup sync.WaitGroup
+	waitGroup.Add(len(progressEventChannels))
+
+	for _, channel := range progressEventChannels {
+		progressResult := &Result{}
+		results = append(results, progressResult)
```
IllyaYalovyy, line 117: It feels like you mixing up together communication be sharing memory (strong anti pattern in Go community) `progressResult` with sharing by communications (channels) - `receiver`. This seems like a fragile approach. I think it requires deeper discussion.

packages/cli/internal/pkg/aws/cdk/progress.go
```diff
@@ -41,19 +50,145 @@ func (p ProgressStream) DisplayProgress(description string) error {
...
+
+func sendDataToReceiverAndUpdateResult(channel <-chan ProgressEvent, progressResult *Result, waitGroup *sync.WaitGroup, receiver chan ProgressEvent) {
+	defer waitGroup.Done()
+
+	var lastEvent ProgressEvent
+	var stopProcessingEvent = ProgressEvent{
+		CurrentStep: 1,
+		TotalSteps:  1,
+	}
+	for initialChannelOut := range channel {
```
IllyaYalovyy, line 134: What does "initialChannelOut" means. I'm not sure I understand the "initial" part.

packages/cli/internal/pkg/aws/cdk/progress.go
```diff
@@ -41,19 +50,145 @@ func (p ProgressStream) DisplayProgress(description string) error {
...
+		if initialChannelOut.Err != nil {
+			progressResult.Err = initialChannelOut.Err
+
+			stopProcessingEvent.UniqueKey = lastEvent.UniqueKey
+			receiver <- stopProcessingEvent
+
+			progressResult.Outputs = lastEvent.Outputs
+			progressResult.UniqueKey = lastEvent.UniqueKey
+			return
```
IllyaYalovyy, line 143: The thing here is a race condition. We can discuss it offline, but in a nutshell, this last "stopProcessing" event can be missed, because waitGroup is released and program exits BEFORE, it has a chance to process the event.

packages/cli/internal/pkg/aws/cdk/deploy_app.go
```diff
@@ -8,7 +8,7 @@ import (
 
 var mkDirTemp = os.MkdirTemp
 
-func (client Client) DeployApp(appDir string, context []string) (ProgressStream, error) {
+func (client Client) DeployApp(appDir string, context []string, uniqueKey string) (ProgressStream, error) {
```
IllyaYalovyy, line 11: This `uniqueKey` argument feels out of place, can we consider other options? 
- use context to propagate this unique key (we can also use context for timeouts)
- auto-generate this unique key

IllyaYalovyy (changes requested):

IllyaYalovyy (commented):
May be it make sense to review some common concurrency patterns:

https://go.dev/blog/pipelines

packages/cli/internal/pkg/aws/cdk/progress.go
```diff
@@ -21,13 +22,21 @@ type ProgressEvent struct {
...
+	Outputs       []string
+	Err           error
 }
 
 func (p ProgressStream) DisplayProgress(description string) error {
 	var lastEvent ProgressEvent
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
-	barReceiver := runProgressBar(ctx, description)
+	barReceiver := runProgressBar(ctx, description, 1)
```
IllyaYalovyy, line 39: This is a bit hard to read. From the code, it's not clear what that magic number "1" means. It doesn't block the PR, but I would much appreciate, if you consider adding a bit of clarity:

```go
...
numberOfChannels := 1
barReceiver := runProgressBar(ctx, description, numberOfChannels)
...
```

packages/cli/internal/pkg/cli/context/interface.go
```diff
@@ -1,9 +1,9 @@
 package context
 
 type Interface interface {
-	Deploy(contextName string, showProgress bool) error
+	Deploy(contexts []string) []ProgressResult
```
IllyaYalovyy, line 4: Does it make sense to still have a single error in this function's result?

IllyaYalovyy (commented):
Just a few minor comments.

packages/cli/internal/pkg/aws/cdk/execute_cdk.go
```diff
@@ -20,17 +20,18 @@ var progressRegex = regexp.MustCompile(`^.*\|\s*([0-9]+/[0-9]+)\s*\|(.*)`)
...
 var osRemoveAll = os.RemoveAll
 
-func executeCdkCommand(appDir string, commandArgs []string) (ProgressStream, error) {
-	return executeCdkCommandAndCleanupDirectory(appDir, commandArgs, "")
+func executeCdkCommand(appDir string, commandArgs []string, executionName string) (ProgressStream, error) {
+	return executeCdkCommandAndCleanupDirectory(appDir, commandArgs, "", executionName)
 }
 
-func executeCdkCommandAndCleanupDirectory(appDir string, commandArgs []string, tmpDir string) (ProgressStream, error) {
+// executionName
```
tneely, line 27: nit: can you remove this comment?

tneely (approved):

IllyaYalovyy (approved):

PR #144, opened 2021-11-04
feat: timeout for context and core deployments

abassyiouni:
**Description of Changes**

Added DeployWithTimeout method, to timeout deployments to 30 minutes. If a deployment times out, an error is displayed to the user asking them to view their cloudformation stack.

- Reduced timeout time to 10 seconds, and ran agc account activate:
```
$ agc account activate
2021-11-04T10:16:31-07:00 𝒊  Activating AGC with bucket '' and VPC ''
Activating account... [____________________________________________________________________________________________________________________________________________________________________________________________]9.8s2021-11-04T10:16:42-07:00 ✘   error="deployment taking longer then expected. please review stack deployment in cloudformation"
Error: an error occurred invoking 'account activate'
with variables: {bucketName: vpcId:}
caused by: deployment taking longer then expected. please review stack deployment in cloudformation
```
- Ran `make` to run tests
- Validated changes don't mess with deployment by running account activate and context deploy, then running workflow

**Checklist**

----

### DIFF
--- a/packages/cli/internal/pkg/cli/context/manager_deploy.go
+++ b/packages/cli/internal/pkg/cli/context/manager_deploy.go
@@ -5,6 +5,7 @@ import (
 	"path/filepath"
 
 	"github.com/aws/amazon-genomics-cli/internal/pkg/aws/cdk"
+	"github.com/aws/amazon-genomics-cli/internal/pkg/cli/util"
 	"github.com/aws/amazon-genomics-cli/internal/pkg/logging"
 	"github.com/rs/zerolog/log"
 )
@@ -20,8 +21,18 @@ func (m *Manager) Deploy(contextName string, showProgress bool) error {
 	m.setTaskContext(contextName)
 	m.clearCdkContext(contextDir)
 	m.setContextEnv(contextName)
-	m.deployContext(contextName, showProgress)
+	m.deployContextWithTimeout(contextName, showProgress)
 	return m.err
+}
+
+func (m *Manager) deployContextWithTimeout(contextName string, showProgress bool) {
+	err := util.DeployWithTimeout(func() error {
+		m.deployContext(contextName, showProgress)
+		return m.err
+	})
+	if m.err == nil && err != nil {
+		m.err = err
+	}
 }
 
 func (m *Manager) clearCdkContext(appDir string) {
--- a/packages/cli/internal/pkg/cli/context/manager_deploy.go
+++ b/packages/cli/internal/pkg/cli/context/manager_deploy.go
@@ -3,6 +3,7 @@ package context
 import (
 	"fmt"
 	"path/filepath"
+	"time"
 
 	"github.com/aws/amazon-genomics-cli/internal/pkg/aws/cdk"
 	"github.com/aws/amazon-genomics-cli/internal/pkg/cli/util"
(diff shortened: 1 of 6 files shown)


packages/cli/internal/pkg/cli/util/timeout.go
```diff
@@ -0,0 +1,25 @@
...
+	completionChannel := make(chan error)
+	go func() {
+		err := timeoutFunction()
+		completionChannel <- err
+	}()
+	select {
+	case err := <-completionChannel:
+		return err
+	case <-time.After(deploymentTimeout):
+		return errors.New("deployment taking longer then expected. please review stack deployment in cloudformation")
```
tneely, line 23: nit: "*the* deployment *is* taking longer than expected. *P*lease review *the* stack in *C*loud*F*ormation"

packages/cli/internal/pkg/cli/context/manager_deploy.go
```diff
@@ -20,10 +21,20 @@ func (m *Manager) Deploy(contextName string, showProgress bool) error {
...
+func (m *Manager) deployContextWithTimeout(contextName string, showProgress bool) {
+	err := util.DeployWithTimeout(func() error {
+		m.deployContext(contextName, showProgress)
+		return m.err
+	})
+	if m.err == nil && err != nil {
+		m.err = err
+	}
+}
+
```
tneely, line 37: Following the pattern in this manager, this should look more like

```go
func (m *Manager) deployContextWithTimeout(contextName string, showProgress bool) {
        if m.err != nil {
		return
	}
	m.err = util.DeployWithTimeout(func() error {
		m.deployContext(contextName, showProgress)
		return m.err
	})
}
```

packages/cli/internal/pkg/cli/util/timeout.go
```diff
@@ -0,0 +1,25 @@
...
+	// channel to mark when a deployment successfully completes
+	completionChannel := make(chan error)
+	go func() {
+		err := timeoutFunction()
+		completionChannel <- err
+	}()
+	select {
+	case err := <-completionChannel:
+		return err
+	case <-time.After(deploymentTimeout):
```
elliot-smith, line 22: We discussed querying Cloudformation to validate the current state of the stack as it could have deployed successfully. Can you please implement that feature since the deployment is async?

abassyiouni: Yup, that should be a pretty simple addition

tneely (dismissed):

packages/cli/internal/pkg/cli/util/timeout.go
```diff
@@ -0,0 +1,25 @@
+package util
+
+import (
+	"errors"
+	"time"
+)
+
+const (
+	timeoutError = "the deployment is taking longer than expected. Please review the stack in CloudFormation"
```
elliot-smith, line 9: We should retrieve the relevant information for the user since we know the stack name and the region it is being deployed to instead of just providing a generic message

packages/cli/internal/pkg/cli/util/timeout.go
```diff
@@ -0,0 +1,34 @@
...
+	"time"
+
+	"github.com/aws/aws-sdk-go-v2/service/cloudformation/types"
+)
+
+var (
+	timeoutError = "the deployment is taking longer than expected. Current deployment status is %s. Please review the stack in CloudFormation"
+)
+
+func DeployWithTimeout(timeoutFunction func() error, deploymentStackStatus func() (types.StackStatus, error), timeoutDuration time.Duration) error {
```
tneely, line 14: I'm worried this function is too specialized and doing too much to be a utility function - maybe make a simplified `runWithTimeout` utility function and wrap it as needed for the CDK deployments

tneely:
After reviewing this code and Elliot's PR I think we're going to have conflicts since Elliot's PR (#139) rewrites a lot of the code you're modifying. Since merging the two will require a rewrite regardless, I'm thinking we should reimagine the timeouts here.

Quick idea - leverage `context.WithTimeout` (see example [here](https://golang.org/src/context/example_test.go)) by passing it into the context manager

e.g.

```go
ctx := context.WithTimeout(context.Background, 30 * time.Minutes)
o.cdkClient.DeployApp(ctx, cdkAppPath, environmentVars)
```

This obviously needs to be reconciled with Elliot's PR, so maybe work with him to determine the best way forward.

IllyaYalovyy:
I think we should use standard means for timeouts. e.g. `context.WithTimeout(...)`

PR #157, opened 2021-11-09
 feat: added new configure format command

anisatahlil:
Issue #, if available:

**Description of Changes**

Implemented a sub command 'format' under 'configure' command. The 'configure format 'command allows a user to set a default format for output display. The default format is then saved in the configuration. When a user does not specify the 'format' flag, the default format in the config is used. On the other hand, if the user specifies the 'format' flag, that value takes precedence over the default format.

This feature allows the user to set and use a default format instead of specifying a format with every command. 

Validation was also added into this feature. The user will be notified if they input an invalid format. 

```
$ agc configure describe
2021-11-10T12:31:42-08:00 𝒊  Reading user specific configuration
CONFIG
FORMAT	text
USER	atahlil@amazon.com	atahlil3HYhj7
$ agc configure format table
2021-11-10T12:32:03-08:00 𝒊  Setting default format to: 'table'
$ agc configure describe
2021-11-10T12:32:06-08:00 𝒊  Reading user specific configuration
CONFIG
FORMAT	table
USER	atahlil@amazon.com	atahlil3HYhj7
$ agc configure format csv
Error: invalid format type. Valid format types are 'text' and 'table'
$ agc configure describe
2021-11-10T12:32:15-08:00 𝒊  Reading user specific configuration
CONFIG
FORMAT	table
USER	atahlil@amazon.com	atahlil3HYhj7
$ agc context list
2021-11-10T12:32:24-08:00 𝒊  Listing contexts.
Name		EngineName
myContext	cromwell
spotCtx		cromwell
$ agc context list --format=text
2021-11-10T12:32:31-08:00 𝒊  Listing contexts.
CONTEXTSUMMARY	cromwell	myContext
CONTEXTSUMMARY	cromwell	spotCtx
$ agc context list --format=csv
invalid format type. Valid format types are 'text' and 'table'
2021-11-10T12:32:35-08:00 𝒊  Listing contexts.
CONTEXTSUMMARY	cromwell	myContext
CONTEXTSUMMARY	cromwell	spotCtx
$ agc context list --format=table
2021-11-10T12:32:44-08:00 𝒊  Listing contexts.
Name		EngineName
myContext	cromwell
spotCtx		cromwell
```

**Checklist**

----


packages/cli/cmd/application/main.go
```diff
@@ -116,13 +141,28 @@ func buildRootCmd() *cobra.Command {
...
 	return cmd
 }
 
+func setFormatter(opts *formatOpts) string {
+	configClient := opts.configClient
+	formatVars := opts.formatVars
+	if formatVars.format == "" {
+		// read the default format from config
+		defaultFormat, err := configClient.GetFormat()
+		formatVars.format = defaultFormat
```
elliot-smith, line 157: We shouldn't set this value if there was an error with the command. Please move it to the else part of the if below. 
It also looks like we might need to set a default value so we could set that the existing default.

packages/cli/cmd/application/main.go
```diff
@@ -116,13 +141,28 @@ func buildRootCmd() *cobra.Command {
...
 	cmd.Flag("docs").Hidden = true
 
 	return cmd
 }
 
+func setFormatter(opts *formatOpts) string {
+	configClient := opts.configClient
+	formatVars := opts.formatVars
+	if formatVars.format == "" {
+		// read the default format from config
```
elliot-smith, line 155: NIT: You don't need this comment here. If you think it's necessary, please update the function name "GetFormat" so that it is more descriptive.

packages/cli/internal/pkg/cli/config/config_client.go
```diff
@@ -135,3 +135,17 @@ func (c Client) GetUserId() (string, error) {
 	userId := userIdFromEmailAddress(userEmailAddress)
 	return userId, nil
 }
+
+func (c Client) SetFormat(format string) error {
+	configData, _ := c.loadFromFile()
```
elliot-smith, line 140: Why are we ignoring the error here?

packages/cli/internal/pkg/cli/config/config_client.go
```diff
@@ -135,3 +135,17 @@ func (c Client) GetUserId() (string, error) {
 	userId := userIdFromEmailAddress(userEmailAddress)
 	return userId, nil
 }
+
+func (c Client) SetFormat(format string) error {
```
elliot-smith, line 139: Can you please test the SetFormat and GetFormat functions?

packages/cli/internal/pkg/cli/config/io.go
```diff
@@ -25,9 +29,16 @@ func fromYaml(filePath string) (Config, error) {
...
 	}
 	var configData Config
+	setDefault(&configData)
 	if err := yaml.Unmarshal(bytes, &configData); err != nil {
 		return Config{}, err
 	}
-
 	return configData, nil
 }
+func setDefault(configData *Config) Config {
```
elliot-smith, line 38: This approach currently has a bug in it where if the yaml file doesn't exist (or has an error), then Config that is returned won't include the required default text value.

You should probably change the function "toYaml" so that it sets a default Config value (since it now has possible defaults) and then override them with values from the config file.

packages/cli/internal/pkg/cli/configure_format.go
```diff
@@ -0,0 +1,75 @@
...
+	return &formatContextOpts{
+		formatContextVars: vars,
+	}, nil
+}
+
+func (o *formatContextOpts) Validate() error {
+	return nil
+}
+
+// Execute returns a context definition for the specified name.
```
elliot-smith, line 34: Please remove this comment as it looks like a copy paste error.

packages/cli/internal/pkg/cli/configure_format.go
```diff
@@ -0,0 +1,75 @@
...
+	return nil
+}
+func BuildConfigureFormatCommand() *cobra.Command {
+	vars := formatContextVars{}
+	cmd := &cobra.Command{
+		Use:   "format output_format",
+		Short: "Sets default format for output display",
+		Args:  cobra.ExactArgs(1),
+		RunE: runCmdE(func(cmd *cobra.Command, args []string) error {
+			if len(args) == 1 {
```
elliot-smith, line 50: The Args comment on line 48 is forcing this behavior although it also provides a bad error message. To stay consistent with the current package, can you please change Args to "cobra.ArbitraryArgs" and then validating the string passed into the user is both the current number and a correct value. You will need to shuffle around some of the logic below though

packages/cli/internal/pkg/cli/configure_format.go
```diff
@@ -0,0 +1,75 @@
...
+				opts, err := newFormatContextOpts(vars)
+				if err != nil {
+					return err
+				}
+				configClient, err := config.NewConfigClient()
+				if err != nil {
+					return clierror.New(configureFormatCommand, vars, err)
+				}
+				opts.configClient = configClient
+				log.Info().Msgf("Setting default format  to: '%s'", opts.format)
```
elliot-smith, line 61: NIT: Extra space in the message

packages/cli/internal/pkg/cli/configure_format.go
anisatahlil: I noticed with some commands like 'configure describe' and 'configure email', we have cobra.ExactArgs() being used. From what I noticed,  cobra.ExactArgs provides an error message notifying the user. What is the benefit in using cobra.ArbitraryArgs in comparison with the cobra.ExactArgs in this case?

elliot-smith: The error messages from Cobra are not customer friendly which is why we have started to move to custom messages.

I.e. The following doesn't tell the user what arg it accepts so they need to run `agc configure --help`
```
$ agc configure email
Error: accepts 1 arg(s), received 0
```
vs custom which tells them what is required
```
$ agc context deploy
Error: either an 'all' flag or a list of contexts must be provided, but not both
```

elliot-smith (approved):

packages/cli/cmd/application/main.go
```diff
@@ -77,15 +101,20 @@ func BuildCommandDocsForHugo(cmd *cobra.Command, dir string) error {
...
 		Example: `
   Displays the help menu for the specified sub-command.
   /code $ agc account --help`,
 		PersistentPreRun: func(cmd *cobra.Command, args []string) {
-			format.SetFormatter(format.FormatterType(vars.format))
+			opts, err := newFormatOpts(formatVars)
+			if err != nil {
+				log.Error().Err(err)
+			}
 			setLoggingLevel()
```
IllyaYalovyy, line 116: This should be the *first* thing we do in this method, in order to guarantee consistent logging.

packages/cli/internal/pkg/cli/config/config.go
```diff
@@ -6,7 +6,10 @@ type User struct {
 	// so nobody would be tempted to change that in the config file and get inconsistency in between email and userid
 	Id string `yaml:"-"`
 }
-
+type Format struct {
```
IllyaYalovyy, line 9: Could you please explain purpose of this struct? What is the benefit of using it versus just having a string field? We can consider creating a type like `type Format string`.

packages/cli/internal/pkg/cli/config/config_client.go
```diff
@@ -135,3 +135,20 @@ func (c Client) GetUserId() (string, error) {
...
+	configData.Format.Format = format
+	return c.storeToFile(configData)
+}
+
+func (c Client) GetFormat() (string, error) {
+	configData, err := c.loadFromFile()
+	if err != nil {
+		return "", err
+	}
+	return configData.Format.Format, nil
```
IllyaYalovyy, line 153: To my previous point. Don't you think this `...Format.Format...` repetition looks and feels weird? It would look better if we have something like `...Format.Name...`, but still not convinced that just simple `configData.Format` will not suffice.

packages/cli/internal/pkg/cli/configure_format.go
```diff
@@ -0,0 +1,77 @@
...
+			if err != nil {
+				return clierror.New(configureFormatCommand, vars, err)
+			}
+			opts.configClient = configClient
+			log.Info().Msgf("Setting default format to: '%s'", opts.format)
+			err = opts.Execute()
+			if err != nil {
+				return clierror.New(configureFormatCommand, vars, err)
+			}
+
```
IllyaYalovyy, line 72: nit. This empty line can be removed

packages/cli/internal/pkg/cli/configure_format_test.go
```diff
@@ -0,0 +1,26 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	textFormat = "text"
+)
+
+func TestFormatContextOpts_Execute(t *testing.T) {
```
IllyaYalovyy, line 13: Can we also test setting some unsupported format? e.g. "csv"

packages/cli/internal/pkg/cli/configure_format.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+func newFormatContextOpts(vars formatContextVars) (*formatContextOpts, error) {
+	return &formatContextOpts{
+		formatContextVars: vars,
+	}, nil
+}
+
+func (o *formatContextOpts) Validate(args []string) error {
+	if len(args) == 0 {
```
IllyaYalovyy, line 33: How do we make sure user provided one of supported formats?

packages/cli/internal/pkg/cli/configure_format.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	return nil
+}
+func BuildConfigureFormatCommand() *cobra.Command {
+	vars := formatContextVars{}
+	cmd := &cobra.Command{
+		Use:   "format output_format",
+		Short: "Sets default format for output display",
+		Args:  cobra.ArbitraryArgs,
```
IllyaYalovyy, line 52: Can we be more specific on what expected?

IllyaYalovyy (changes requested):

packages/cli/internal/pkg/cli/config/config.go
anisatahlil: The way the output displays seems to rely on the struct structure. By just adding format as string field in the Config struct, the 'configure describe' command yields an output like this.  
```
CONFIG	text
USER	atahlil@amazon.com	atahlil3HYhj7
```

Where as adding the struct would yield an output like this 
```
CONFIG	
FORMAT    text
USER	atahlil@amazon.com	atahlil3HYhj7
```
I am not sure if there is a better way to construct this without compromising the way that the output will display

packages/cli/internal/pkg/cli/config/config_client.go
anisatahlil: Agreed. I will update that to something more readable

packages/cli/internal/pkg/cli/config/config.go
```diff
@@ -6,7 +6,10 @@ type User struct {
...
+	Value string `yaml:"format"`
+}
 type Config struct {
-	User User `yaml:"user"`
+	User   User `yaml:"user"`
+	Format Format
```
abassyiouni, line 14: should this be `yaml:"format"`

abassyiouni, line 14: should this have yaml:"format"

abassyiouni (approved):

packages/cli/cmd/application/main.go
```diff
@@ -77,15 +101,20 @@ func BuildCommandDocsForHugo(cmd *cobra.Command, dir string) error {
...
   Displays the help menu for the specified sub-command.
   /code $ agc account --help`,
 		PersistentPreRun: func(cmd *cobra.Command, args []string) {
-			format.SetFormatter(format.FormatterType(vars.format))
 			setLoggingLevel()
+			opts, err := newFormatOpts(formatVars)
+			if err != nil {
+				log.Error().Err(err)
+			}
+			setFormatter(opts)
```
IllyaYalovyy, line 117: Did you notice what happens when "newFormatOpts" fails? in this case setFormatter will be called with empty opts. `storage.ConfigClient` is an interface which is expected to be populated with a pointer to an actual client implementation. In this case it will be `nil` and any attempt to deference this pointer will cause a crash. It would be nice to cover this in unit tests.

packages/cli/cmd/application/main.go
```diff
@@ -116,12 +145,37 @@ func buildRootCmd() *cobra.Command {
...
 
 	cmd.PersistentFlags().BoolVarP(&logging.Verbose, cli.VerboseFlag, cli.VerboseFlagShort, false, cli.VerboseFlagDescription)
-	cmd.PersistentFlags().StringVar(&vars.format, cli.FormatFlag, cli.FormatFlagDefault, cli.FormatFlagDescription)
+	cmd.PersistentFlags().StringVar(&formatVars.format, cli.FormatFlag, "", cli.FormatFlagDescription)
 	cmd.Flags().StringVar(&vars.docPath, "docs", "", "generate markdown documenting the CLI to the specified path")
 	cmd.Flag("docs").Hidden = true
 
 	return cmd
 }
+func ValidateFormat(f format.FormatterType) error {
```
IllyaYalovyy, line 154: Could you please add empty lines between methods, we want our code base to be consistent.

packages/cli/cmd/application/main.go
```diff
@@ -116,12 +145,37 @@ func buildRootCmd() *cobra.Command {
...
+	if err := f.ValidateFormatter(); err != nil {
+		return err
+	}
+	return nil
+}
+func setFormatter(opts *formatOpts) string {
```
IllyaYalovyy, line 160: Why "setter"  returns a value? I don't see it used in the main code.

packages/cli/cmd/application/main.go
```diff
@@ -77,15 +101,20 @@ func BuildCommandDocsForHugo(cmd *cobra.Command, dir string) error {
...
 	cmd := &cobra.Command{
 		Use:   "agc",
 		Short: shortDescription,
 		Example: `
   Displays the help menu for the specified sub-command.
   /code $ agc account --help`,
 		PersistentPreRun: func(cmd *cobra.Command, args []string) {
-			format.SetFormatter(format.FormatterType(vars.format))
 			setLoggingLevel()
+			opts, err := newFormatOpts(formatVars)
```
IllyaYalovyy, line 113: - It doesn't seem like error affect our execution flow in any way here. It means we can make this code simpler:
```go
PersistentPreRun: func(cmd *cobra.Command, args []string) {
  setLoggingLevel()
  setFormatter(formatVars)
  checkCliVersion()
```
In my opinion this way the lambda looks cleaner without loosing any functionality.

Creation of a client and logging of the error can happen inside `setFormatter` method itself.

```go
func setFormatter(formatVars formatVars) {
	configClient, err := config.NewConfigClient()
	if err != nil {
		log.Error().Err(err)
		return
	}
...

}
```

In order to make it testable, you can use a factory method for client creation.

```go
var newConfigClient = config.NewConfigClient
```

packages/cli/cmd/application/set_format.go
```diff
@@ -0,0 +1,37 @@
...
+		}
+	}
+	if err := ValidateFormat(format.FormatterType(f.format)); err != nil {
+		fmt.Println(err.Error())
+	}
+	format.SetFormatter(format.FormatterType(f.format))
+	return f.format
+}
+
+func ValidateFormat(f format.FormatterType) error {
```
abassyiouni, line 32: Is this required?

anisatahlil: yeah I can update that to use the ValidateFormatter directly

IllyaYalovyy (approved):

elliot-smith (approved):

PR #167, opened 2021-11-12
fix: use proper assertion in order to avoid sporadic unit test failures

IllyaYalovyy:
Issue #, if available:

**Description of Changes**

Events are accumulated in a map with key being log stream name. Order of keys in map is not guaranteed by the implementation. In unit test we cannot require particular order, or it will cause sporadic unit test failures. 

- make
- unit test is back to stable

**Checklist**

----

### DIFF
--- a/packages/cli/internal/pkg/aws/cwl/stream_logs_test.go
+++ b/packages/cli/internal/pkg/aws/cwl/stream_logs_test.go
@@ -53,15 +53,21 @@ func TestClient_StreamLogs(t *testing.T) {
 			}}, nil)
 	cancel()
 	stream := client.StreamLogs(ctx, testLogGroupName)
+	logStream1Event1 := fmt.Sprintf("%s\tHello", someTime1.Format(time.RFC1123Z))
+	logStream1Event2 := fmt.Sprintf("%s\tworld!", someTime2.Format(time.RFC1123Z))
+	logStream2Event1 := fmt.Sprintf("%s\tHola", someTime1.Format(time.RFC1123Z))
+	logStream2Event2 := fmt.Sprintf("%s\tmundo!", someTime2.Format(time.RFC1123Z))
 	event := <-stream
-	assert.ElementsMatch(t, []string{
-		fmt.Sprintf("%s\tHello", someTime1.Format(time.RFC1123Z)),
-		fmt.Sprintf("%s\tworld!", someTime2.Format(time.RFC1123Z)),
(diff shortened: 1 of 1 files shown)


packages/cli/internal/pkg/aws/cwl/stream_logs_test.go
```diff
@@ -54,7 +54,7 @@ func TestClient_StreamLogs(t *testing.T) {
 	cancel()
 	stream := client.StreamLogs(ctx, testLogGroupName)
 	event := <-stream
-	assert.Equal(t, []string{
+	assert.ElementsMatch(t, []string{
```
elliot-smith, line 57: The order is somewhat important. Do you mind asserting that the values at 1 and 3 are either Hello's? There is an example here: https://github.com/aws/amazon-genomics-cli/blob/main/packages/cli/internal/pkg/aws/cwl/get_logs_test.go#L92

IllyaYalovyy: if order is important we should not use map, or if we use a map we should reorder it afterwards. In this case we want event that belong to the same log stream to be ordered properly. I don't thing we can guarantee concrete index of the event.

tneely (approved):

ghawk1ns (dismissed):

packages/cli/internal/pkg/aws/cwl/stream_logs_test.go
elliot-smith: That is what the test is doing. If you look at the errors that are associated with the test it is because the "Hola" is being shown first sometimes. The only order that matters is that "Hello" and "World" are next to each other and "Hola" is next to "Mundo". If there is an easier way to validate the ordering then please fix it.

If this test printed out "Hello", "Hola", "Mundo", "World" then it would be invalid but the change that you are making makes it valid.

elliot-smith (approved):

packages/cli/internal/pkg/aws/cwl/stream_logs_test.go
IllyaYalovyy: Will it be correct if it prints "World", "Hello"?

elliot-smith (approved):

packages/cli/internal/pkg/aws/cwl/stream_logs_test.go
elliot-smith: We really should test that as well, good catch.

tneely (approved):

PR #189, opened 2021-11-23
perf: Increased RAM allocation for miniwdl engine

markjschreiber:
Issue #, if available:

**Description of Changes**

Increased memory allocation for miniwdl to 4096 MiB

* `make test`
* `agc context deploy miniwdlCtx`
* Run a workflow.
* Verify memory allocation of generated job definition in console.

**Checklist**

- [*] I have linted my code before raising the PR
- [*] Title of this Pull Request follows Conventional Commits standard: https://www.conventionalcommits.org/en/v1.0.0/

----

### DIFF
--- a/packages/cdk/lib/constructs/engines/miniwdl/miniwdl-engine.ts
+++ b/packages/cdk/lib/constructs/engines/miniwdl/miniwdl-engine.ts
@@ -18,6 +18,7 @@ const MINIWDL_IMAGE_DESIGNATION = "miniwdl";
 export class MiniWdlEngine extends Engine {
   readonly headJobDefinition: JobDefinition;
   private readonly volumeName = "efs";
+  private readonly engineMemoryMiB = 4096;
 
   constructor(scope: Construct, id: string, props: MiniWdlEngineProps) {
     super(scope, id);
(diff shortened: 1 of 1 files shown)


AndreyDovydenko (approved):

packages/cdk/lib/constructs/engines/miniwdl/miniwdl-engine.ts
```diff
@@ -34,6 +35,7 @@ export class MiniWdlEngine extends Engine {
       logGroup: this.logGroup,
       platformCapabilities: [PlatformCapabilities.FARGATE],
       container: {
+        memoryLimitMiB: this.engineMemoryMiB,
```
AndreyDovydenko, line 38: Is there any cost implications out of this change that we need to consider here, looks like something that we could have as a parameter in our context definition.

markjschreiber: There are although the engine only runs for the lifetime of the workflow so there are not "ongoing costs". Rather than make this configurable I would rather work with Mike Lin to see if we can reduce the need for large memory for workflows with large scatters.

tneely (approved):

PR #197, opened 2021-11-30
fix: returning default config with initial email config

anisatahlil:
Issue #, if available:

**Description of Changes**
Bug when email is initially configured using `agc email configure`. The format was not initialized properly causing an error on every agc command

```
$ agc configure email test@amazon.com
2021-11-30T09:48:38-08:00 𝒊  Setting user email address to: 'test@amazon.com'
$ agc configure email test@amazon.com
2021-11-30T09:49:10-08:00 𝒊  Setting user email address to: 'test@amazon.com'

```
config file 
```
user:
    email: test@amazon.com
format:
    name: ""
```
```
$ agc context list
invalid format type. Valid format types are 'text' and 'table'
2021-11-30T09:49:20-08:00 𝒊  Listing contexts.
CONTEXTSUMMARY	cromwell	ctx1
$ agc configure describe
invalid format type. Valid format types are 'text' and 'table'
2021-11-30T09:49:32-08:00 𝒊  Reading user specific configuration
CONFIG
FORMAT
USER	test@amazon.com	test3HYhj7
```

Validates fix by running the same command

```
$ agc configure email test@amazon.com
2021-11-30T09:53:27-08:00 𝒊  Setting user email address to: 'test@amazon.com'
$ agc context list
2021-11-30T09:53:38-08:00 𝒊  Listing contexts.
CONTEXTSUMMARY	cromwell	ctx1
```
config file
```
user:
    email: test@amazon.com
format:
    name: text
```

**Checklist**

----

### DIFF
--- a/packages/cli/internal/pkg/cli/config/config_client.go
+++ b/packages/cli/internal/pkg/cli/config/config_client.go
@@ -76,7 +76,7 @@ func (c Client) Read() (Config, error) {
 func (c Client) loadFromFile() (Config, error) {
 	configData, err := fromYaml(c.configFilePath)
 	if err != nil {
-		return Config{}, err
+		return configData, err
 	}
 	configData.User.Id = userIdFromEmailAddress(configData.User.Email)
 	return configData, nil
--- a/packages/cli/internal/pkg/cli/config/config_test.go
+++ b/packages/cli/internal/pkg/cli/config/config_test.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"errors"
 	"io/fs"
 	"testing"
 
@@ -29,6 +30,15 @@ var (
 			Name: "text",
 		},
 	}
+
+	expectedDefaultConfig = Config{
+		User{
+			Email: "",
+		},
+		Format{
+			Name: "text",
+		},
+	}
 )
 
 func TestConfig_ReadData(t *testing.T) {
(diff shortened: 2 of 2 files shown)


packages/cli/internal/pkg/cli/config/config_client.go
```diff
@@ -76,7 +76,7 @@ func (c Client) Read() (Config, error) {
 func (c Client) loadFromFile() (Config, error) {
 	configData, err := fromYaml(c.configFilePath)
 	if err != nil {
-		return Config{}, err
+		return configData, err
```
AndreyDovydenko, line 79: When we enter this condition statement our `fromYaml` call has returned an error reading data. Could we trust the data object returned from it and pass it out?

elliot-smith: This is the current fromYaml function which returns back the default values

```
func fromYaml(filePath string) (Config, error) {
	configData := Config{Format: Format{defaultFormat}}
	bytes, err := readFile(filePath)
	if err != nil {
		return configData, err
	}
	if err := yaml.Unmarshal(bytes, &configData); err != nil {
		return configData, err
	}
	return configData, nil
}

```

AndreyDovydenko: Are we sure setting default format in this function and reading data from it when the function returns with a none nil error code is the best approach here?

`fromYaml` name makes me think that the responsibility of the function is to read the data from a yaml file. If we want to have a `fallback default value` should it live in a proper abstraction level function? E.g. 'try to read fromYaml() if failed return defaultConfig`. Another option that we could take is to try to repair / generate default config file and then read the correct file again. I'm not convinced we should be eating up an error and trusting the data returned from a function giving us an error when we call it.

elliot-smith: I tend to agree with this as it does seem like a weird pattern especially since it doesn't follow Go patterns. I would prefer it to be fixed in a follow up PR though as there will be other code that is impacted and I would like the code changes to be small and isolated if possible.

AndreyDovydenko: is anybody calling into `fromYaml` beside us in this function?

anisatahlil: `fromYaml` is used throughout whenever config file is being read. I understand the concern here. Would a better approach be to create a function in this file that handles the setting of the default config. That way, if `fromYaml` fails and returns an error , then the function can be called which sets the default config.

AndreyDovydenko: it looks like most of the other functions call into `loadFromFile` and not `fromYaml` directly though? I think yes, having a function that narrowly handles adding / setting defaults is a good idea.

anisatahlil: ahh yes, seem to have misunderstood the question but yes it is only being called from `loadFromFile`. 
Sounds good. I will add a follow up PR to address this

AndreyDovydenko (approved):

anisatahlil:
Closing this pull request as the change is being addressed in https://github.com/aws/amazon-genomics-cli/pull/201

PR #198, opened 2021-11-30
fix: fixed arg validation for format

anisatahlil:
Issue #, if available:

**Description of Changes**
Argument checks for `agc configure format` was not handled gracefully. This caused stack error to output into the console.
```
$ agc configure format
panic: runtime error: index out of range [0] with length 0
```

```
$ agc configure format
Error: format value must be provided
```

**Checklist**

----

### DIFF
--- a/packages/cli/internal/pkg/cli/configure_format.go
+++ b/packages/cli/internal/pkg/cli/configure_format.go
@@ -30,8 +30,8 @@ func newFormatContextOpts(vars formatContextVars) (*formatContextOpts, error) {
 	}, nil
 }
 
-func (o *formatContextOpts) Validate() error {
-	if o.formatContextVars.format == "" {
+func (o *formatContextOpts) Validate(args []string) error {
+	if len(args) == 0 {
 		return fmt.Errorf("format value must be provided")
 	}
 	format := format.FormatterType(o.formatContextVars.format)
@@ -61,7 +61,7 @@ func BuildConfigureFormatCommand() *cobra.Command {
 			if err != nil {
 				return err
 			}
-			if err := opts.Validate(); err != nil {
+			if err := opts.Validate(args); err != nil {
 				return err
 			}
 			configClient, err := config.NewConfigClient()
--- a/packages/cli/internal/pkg/cli/configure_format.go
+++ b/packages/cli/internal/pkg/cli/configure_format.go
@@ -31,8 +31,8 @@ func newFormatContextOpts(vars formatContextVars) (*formatContextOpts, error) {
 }
 
 func (o *formatContextOpts) Validate(args []string) error {
-	if len(args) == 0 {
-		return fmt.Errorf("format value must be provided")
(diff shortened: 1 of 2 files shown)


packages/cli/internal/pkg/cli/configure_format.go
```diff
@@ -56,14 +56,14 @@ func BuildConfigureFormatCommand() *cobra.Command {
 		Short: "Sets default format option for output display of AGC commands. Valid format options are 'text' and 'table'",
 		Args:  cobra.ArbitraryArgs,
```
AndreyDovydenko, line 57: Should this be `Args:  cobra.ExactArgs(1),` to ensure the single argument passed. or is there a case where we would want to have more than one arg passed to this command?

packages/cli/internal/pkg/cli/configure_format.go
```diff
@@ -30,8 +30,8 @@ func newFormatContextOpts(vars formatContextVars) (*formatContextOpts, error) {
 	}, nil
 }
 
-func (o *formatContextOpts) Validate() error {
-	if o.formatContextVars.format == "" {
+func (o *formatContextOpts) Validate(args []string) error {
```
AndreyDovydenko, line 33: Should we be passing an array in here if we're only reading `args[0]` out of it later on?

AndreyDovydenko (changes requested):

packages/cli/internal/pkg/cli/configure_format.go
anisatahlil: The feedback from pervious PRs suggested to move away from `Args: cobra.ExactArgs(1)` to give customers a more readable error when argument is not provided.

packages/cli/internal/pkg/cli/configure_format.go
anisatahlil: That function is handling validation which also checks if the array is empty to provide users with a readable error message

packages/cli/internal/pkg/cli/configure_format.go
```diff
@@ -30,8 +30,8 @@ func newFormatContextOpts(vars formatContextVars) (*formatContextOpts, error) {
...
+	if len(args) == 0 {
```
elliot-smith, line 34: This should be != 1 with a message similar to "a single format value must be provided"

packages/cli/internal/pkg/cli/configure_format.go
elliot-smith: "Error: accepts 1 arg(s), received 0" only tells the customer the what while "Error: a single format value must be provided" tells the customer the why as well. This means that for a customer to resolve the issue, they don't need to type "-h" to know what the values are.

Ideally, we could populated the valid args like you mention or generate the message with the valid options like we do with some of our other commands.

elliot-smith (approved):

PR #201, opened 2021-12-01
fix: function for default config

anisatahlil:
Issue #, if available:

**Description of Changes**
This is a follow up to address the comments in https://github.com/aws/amazon-genomics-cli/pull/197. This change involves adding a function that returns the default config.

``` 
$ agc configure email test@email.com
2021-12-01T09:29:09-08:00 𝒊  Setting user email address to: 'test@email.com'
$ agc configure describe
2021-12-01T09:29:12-08:00 𝒊  Reading user specific configuration
CONFIG
FORMAT  text
USER    test@email.com  test1X6OLM
$ agc 
```

**Checklist**

----

### DIFF
--- a/packages/cli/internal/pkg/cli/config/config_client.go
+++ b/packages/cli/internal/pkg/cli/config/config_client.go
@@ -16,6 +16,8 @@ const (
 	configFileName = "config.yaml"
 )
 
+var defaultConfig = Config{Format: Format{defaultFormat}}
+
 type Client struct {
 	configFilePath string
 }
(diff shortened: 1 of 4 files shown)


packages/cli/internal/pkg/cli/config/config_client.go
```diff
@@ -68,15 +68,17 @@ func userIdFromEmailAddress(emailAddress string) string {
 	sanitizedUserName := sanitizeUserName(userName)
 	return sanitizedUserName + hash(emailAddress)
 }
-
+func defaultConfig() Config {
```
elliot-smith, line 71: This can just be a variable

packages/cli/internal/pkg/cli/config/io.go
```diff
@@ -24,13 +24,13 @@ func toYaml(filePath string, configData Config) error {
 }
 
 func fromYaml(filePath string) (Config, error) {
```
elliot-smith, line 26: The more I read this the less I like it due to the tight coupling to the Config interface. Ideally, this should use Generics which will be released with Go 1.18 in February in 2022.

If we wanted this to read like Go 1.17 code, we should pass in the interface to "fromYaml" which then passes it through to the yaml.Unmarhsal function. This would allow the fromYaml function to be generic as the function name suggests that it is. It isn't a large change so despite it being pre-optimization this is probably the correct path forward.

anisatahlil: Yeah fromYaml is tightly coupled with the config Interface and generics would help resolve that once that is released. I am okay to pass the interface to set this function on a path to work more generically. But I am also wondering if we have similar functions that can benefit from generics, would need similar changes and if that could perhaps be addressed in a separate PR

elliot-smith: We would have other code like that but lets keep it simple for now and just address this file for now and address the others as they come up.

AndreyDovydenko (approved):

elliot-smith (approved):

PR #206, opened 2021-12-03
refactor: Generates the ECR images generically

elliot-smith:
**Description of Changes**
We now support multiple images inside of ECR, the text is generic and the file is becoming fairly long. This change refactors the ECR images to be referenced as part of the context step instead of the activate step. This method keeps the existing values the same so if a customer wants to override the values, they can use the existing values.

NB: Once this has been approved, I will update the documentation as an amendment to this PR.

Ran `agc account activate`, `agc context deploy --all`, `agc context destroy --all` and `agc account deactivate`. I also override the region for one of the core values and it was overridden correctly.

**Checklist**

----


packages/cdk/apps/core/app.ts
```diff
@@ -13,6 +14,43 @@ const region: string = process.env.CDK_DEFAULT_REGION!;
...
+        value: getContext(app.node, `ECR_${ecrImage.name.toUpperCase()}_TAG`),
+        description: `Docker tag for the ${ecrImage.descriptionName} image`,
+      },
+      {
+        name: `${ecrImage.name}/ecr-repo/repository`,
+        value: getContext(app.node, `ECR_${ecrImage.name.toUpperCase()}_REPOSITORY`),
+        description: `ECR repository for the ${ecrImage.descriptionName} image`,
+      }
+    );
+  }
```
nbraid, line 50: IIRC this is where we set the SSM parameters on `account activate` that are then read when we `context deploy` something that needs some of these images?

Do we need SSM at all for this? Perhaps this is a bigger refactor that should be done in a separate task, but I wonder if we could just code these repositories and accounts into the Context stack, overridable by a context parameter there if we so needed. `agc context deploy myContext --cdk-args "-c ECR_MINIWDL_REPOSITORY=agc/miniwdl-mirror-nightly"`

Concretely, this would remove a reason to have to `account activate` when upgrading AGC. To pick up a new version of a WDL or WES image, you just need to redeploy the context that you want to upgrade.

elliot-smith: The one downside that I can see with that approach is that if you re-deploy a context then you are changing the underlying images that are being used by AGC to run your workflows which might cause issues in the future if a new version of an engine breaks compatibility and would require users to modify how they interact with AGC (which was why I thought that it was implemented this way to begin with).

I'll dig a little deeper to see if we can make your change.

tneely: The SSM parameters were a relic of an older design, and we've discussed removing them before though we never got around to actually doing it.

Regarding breaking compatibility, we have some protections around destroying a context if there are running workflows. We could extend that to redeploying the context as well

PR #208, opened 2021-12-06
refactor: Changes the ECR images to be generic and referenced during context deploy

elliot-smith:
**Description of Changes**

We now support multiple images inside of ECR that is hosted in an AWS account, the text is generically defined and the file that defines them is becoming long. This change refactors the ECR images to be referenced as part of the context deploy step instead of the activate step as well as allowing the images to be updated in one place (instead of several). This method keeps the existing values the same so if a customer wants to override the values, they can use the existing values.

Ran agc account activate, agc context deploy --all, agc context destroy --all and agc account deactivate. I also override the region for one of the core values and it was overridden correctly.

**Checklist**

----


packages/cli/internal/pkg/cli/context/manager.go
```diff
@@ -219,6 +241,24 @@ func (m *Manager) setContextEnv(contextName string) {
...
+		m.err = actionableerror.New(
+			fmt.Errorf("the engine name in your context file '%s' does not exist", m.contextEnv.EngineName),
+			"Please check your agc config file for the engine you have supplied",
+		)
+		return
+	}
+
+	if err := m.ecrClient.VerifyImageExists(imageRef); err != nil {
+		m.err = err
+	}
```
tneely, line 260: We can simplify this to `m.err = m.ecrClient.VerifyImageExists(imageRef)`

packages/cli/internal/pkg/cli/context/manager.go
```diff
@@ -78,6 +83,21 @@ type ProgressResult struct {
...
 
+func (m *Manager) getEnvironmentVars() []string {
+	var environmentVars []string
+
+	for imageName := range m.imageRefs {
+		environmentVars = append(environmentVars,
+			fmt.Sprintf("ECR_%s_ACCOUNT_ID=%s", imageName, m.imageRefs[environment.WesImageKey].RegistryId),
+			fmt.Sprintf("ECR_%s_REGION=%s", imageName, m.imageRefs[environment.WesImageKey].Region),
+			fmt.Sprintf("ECR_%s_TAG=%s", imageName, m.imageRefs[environment.WesImageKey].ImageTag),
+			fmt.Sprintf("ECR_%s_REPOSITORY=%s", imageName, m.imageRefs[environment.WesImageKey].RepositoryName),
```
tneely, line 94: Why are these always the WesImageKey?

tneely (changes requested):

packages/cli/internal/pkg/cli/context/manager.go
elliot-smith: Good catch, I've also updated the underlying images so that our tests don't always use "WES"

tneely (approved):

AndreyDovydenko (approved):

PR #230, opened 2021-12-15
fix: Scope down batch role permissions

tneely:
**Description of Changes**

Our EBS autoscaling role allows Batch jobs to delete any EBS volume. Because we also create the volumes as needed, we can't scope down to the resource level. Since volumes are tagged by the same tags as the instance, we can however restrict the delete policy to just volumes with the `agc` tag. The role also has tag permissions, so we can't protect users that run malicious over code, but we can still safeguard against accidents and reduce the blast radius to just the account and region the user is operating in.

I also refactored a file into `lib`

I ran the hello and haplotype example tests for the cromwell and miniwdl contexts before and after the change. Both succeeded. Confirmed EBS volumes are destroyed after the job ends.

**Checklist**

----


packages/cdk/lib/constructs/batch.ts
```diff
@@ -124,8 +125,21 @@ export class Batch extends Construct {
         "ebs-autoscaling": new PolicyDocument({
           statements: [
             new PolicyStatement({
-              actions: ["ec2:CreateTags", "ec2:DescribeVolumes", "ec2:CreateVolume", "ec2:AttachVolume", "ec2:DeleteVolume", "ec2:ModifyInstanceAttribute"],
-              resources: ["*"],
+              actions: ["ec2:DescribeVolumes", "ec2:CreateVolume", "ec2:CreateTags"],
+              resources: [`arn:aws:ec2:${Aws.REGION}:${Aws.ACCOUNT_ID}:volume/*`],
```
nbraid, line 129: I created a dumb little [util function](https://github.com/aws/amazon-genomics-cli/blob/main/packages/cdk/lib/util/index.ts#L110-L112) for doing something similar with Batch. 

I'm fine with leaving it as-is, but if you think adding a `service` parameter or make a similar function for EC2 ARNs would be more readable then go for it.

nbraid (approved):

elliot-smith (approved):

packages/cdk/lib/constructs/batch.ts
tneely: That's a cleaner approach - I'll add an `ec2Arn` function

nbraid (approved):

elliot-smith (approved):

nbraid (approved):

Issue #124, opened 2021-10-27
Doesn't work with accounts using MFA


adamnovak:
[//]: #  (Please do not include your AWS account or credentials as part of the bug request.)

**Describe the Bug**

[//]: #  (Please describe the behavior that you are seeing with as many details as possible including logs from AGC, CloudWatch snippets and CloudFormation errors if relevant.)

I can't activate AGC on my AWS account because the tool can't work with the MFA assume role setup that we use for access control at the UCSC Genomics Institute.

**Steps to Reproduce**

[//]: #  (Include a list of steps that can be used to reproduce the bug.)

1. Set up your AWS accounts. Have one AWS account where all the users are, and get an access key and secret key for your user, and put them in `~/.aws/credentials`. Have another AWS account that is the one you actually want to run AGC in. Make a role there to be assumed, and set up the IAM permissions so that users in the first account can assume the role in the second account when authenticating with MFA. Get the ARN for the role.
2. Set up an MFA device on your IAM user in the first account and get its serial.
3. Write a `~/.aws/config` file that looks something like this, defining a profile that assumes the role in the destination account:
```
[default]
region = us-west-2

[profile toil]
region = us-west-2
role_arn = PASTE_ARN_HERE
source_profile = default
mfa_serial = PASTE_SERIAL_HERE
```
4. Try and set up AGC using the profile you defined:
```
agc account -v -p toil activate
```

**Relevant Logs**

[//]: #  (The logs that where obtained by running the commands here. Please run the command with "-v" so that we can see the verbose logs.)

```
$ agc account -v -p toil activate
2021-10-27T09:48:17-07:00 ☠️   error="assume role with MFA enabled, but AssumeRoleTokenProvider session option not set."
```

**Expected Behavior**

[//]: #  (A clear and concise description of what you expected to happen.)

The activate command should succeed, or else complain that the role doesn't have the permissions it needs to do the setup.

**Actual Behavior**

[//]: #  (A clear and concise description of what actually happened.)

Instead, AGC can't even assume the role; it looks like the piece of the Go SDK that is responsible for interacting with the keyboard and prompting for an MFA code is not set up. Interestingly, the exact same thing happens even when I have an active cached MFA session for the `aws` Python-based tool, so AGC isn't even capable of finding and using the cached assumed-role credentials in `~/.aws/cli/cache/`.

**Screenshots**

[//]: #  (If applicable, add screenshots to help explain your problem.)

Not applicable; everything is text.

[...]

PR #231, opened 2021-12-15
fix: Use token provider for MFA profiles

adamnovak:
Issue #, if available: #124

**Description of Changes**

This adds an MFA token provider to the AWS SDK configuration that gets used, so it can actually prompt for MFA tokens.

I've run `make build` and then `./scripts/run-dev.sh`, and entered `account activate` into it. AGC prompted me for an MFA code and then apparently successfully activated my account.

**Checklist**

  ~I don't have control over the CI to set up an account that actually uses MFA, so I don't think I can do this.~
  This turned out to be `make format`.

----


elliot-smith:
Hi Adam,
Apologies for the slow review on this, we use both CDK and the CLI to interact with AWS resources and want to be sure that this supports both types of interactions as part of your change.

adamnovak:
This only addresses the CLI side, which is the issue I ran into first since it is required for `agc account activate`. I haven't yet tried to deploy any CDK constructs, so I don't know if there are any additional bugs in `aws-cdk-lib` around prompting for MFA credentials that need to be fixed. This repo's CDK constructs don't appear to have any responsibilities around user authentication at all, so if there is an MFA bug in using the CDK it would have to be in the upstream package. 

It looks like `aws-cdk-lib` [defaults to using its AWS-SDK-compatible credentials loader](https://github.com/aws/aws-cdk/blob/69f5f3a8a4447cba5ae28ab8d9ea58d15332e465/packages/aws-cdk/bin/cdk.ts#L212), and that in turn [implements the MFA prompt](https://github.com/aws/aws-cdk/blob/332ce4d9ae995bd1336fef13e2c7f9fc0c12f34d/packages/aws-cdk/lib/api/aws-auth/awscli-compatible.ts#L330-L347), so unless AGC is doing something weird to the CDK library's authentication logic, it can probably be relied upon to actually support MFA.

adamnovak:
OK, I'm going further into the setup process, and the MFA still seems to be working properly.

When I get to the CDK bootstrap step, the `cdk` tool prompts me for MFA already, so I think it is working properly.

```
$ cd $HOME/.agc/cdk
$ npx cdk bootstrap aws://318423852362/us-west-2
 ⏳  Bootstrapping environment aws://318423852362/us-west-2...
MFA token for arn:aws:iam::652235167018:mfa/anovak@soe.ucsc.edu: 273937
CDKToolkit: creating CloudFormation changeset...




 ✅  Environment aws://318423852362/us-west-2 bootstrapped.
****************************************************
*** Newer version of CDK is available [2.3.0]    ***
*** Upgrade recommended (npm install -g aws-cdk) ***
****************************************************
```

adamnovak:
OK, while the CDK does in fact prompt for MFA, when `agc account activate` *invokes* the `cdk` command behind the scenes, the CDK itself prompts for MFA, and AGC doesn't handle this properly.

The output stream from the CDK (where the prompt goes) is being swallowed by AGC unless you pass `-v`. And `cdk` stops immediately after prompting for MFA (because its input stream isn't hooked up?), and AGC sees that it stopped and decides it succeeded and exist with 0, even though the account is not actually activated.

So some more work needs to be done around how `cdk` is called to deal with the CDK needing to interact with the user itself.

adamnovak:
Here's a log:

```
2021-12-22T12:12:22-08:00 ↓  Checking AGC version...
2021-12-22T12:12:22-08:00 ↓  version check failed: assume role with MFA enabled, but AssumeRoleTokenProvider session option not set.
2021-12-22T12:12:22-08:00 𝒊  Activating AGC with bucket '' and VPC ''
Assume Role MFA token code: 123846
2021-12-22T12:12:35-08:00 ↓  executeCDKCommand(/public/home/anovak/.agc/cdk/apps/core, [deploy --all --profile  --require-approval never --output /public/home/anovak/.agc/cdk/apps/core/cdk-output4141619014 -c AGC_BUCKET_NAME=agc-318423852362-us-west-2 -c CREATE_AGC_BUCKET=true])
2021-12-22T12:12:37-08:00 ↓  
2021-12-22T12:12:37-08:00 ↓  > cdk@0.1.0 cdk
2021-12-22T12:12:37-08:00 ↓  > cd $INIT_CWD && cdk "deploy" "--all" "--profile" "" "--require-approval" "never" "--output" "/public/home/anovak/.agc/cdk/apps/core/cdk-output4141619014" "-c" "AGC_BUCKET_NAME=agc-318423852362-us-west-2" "-c" "CREATE_AGC_BUCKET=true"
2021-12-22T12:12:37-08:00 ↓  
2021-12-22T12:12:40-08:00 ↓  MFA token for arn:aws:iam::652235167018:mfa/anovak@soe.ucsc.edu: 
```

adamnovak:
Looks like the config for the version check is being made here: https://github.com/aws/amazon-genomics-cli/blob/8dbcd9d102f96afd0bb7aad556233cbabef921da/packages/cli/internal/pkg/version/check_func.go#L51-L57

Shouldn't the version check *really* just be an unsigned request? I'm going to add MFA to that config, but it seems silly to really authenticate the user to ask for this public version info. It also looks like it uses the default profile all the time, rather than any profile the user has specified on the command line, which isn't quite right.

adamnovak:
I'm having some trouble sorting out how to actually forward along standard input to the child process.

Go doesn't seem to just let it inherit the parent process's standard input file descriptor; I *have* to call it with a standard input pipe.

I have a loop to copy byte by byte from our standard input to the subprocess's standard input, but then I'm never seeing any *output* from the subprocess; it might be being buffered until the subprocess is done?

I'm thinking of refactoring the input-copy to use a channel, because I've heard that Go's IO is actually always blocking and you need to use a channel and a timeout if you want to not block. But I'm not sure that that will actually help; all the IO shoveling is happening in different goroutines already; if one can block all the others on a read, won't that still happen if I have a channel and a timeout?

As is, I have to manually kill the Node process, and then type some input, and the input shoveler will complain about `write |1: broken pipe` (even though I catch and ignore `io.ErrClosedPipe` since I expect it), and *then* after a few moments it will get around to seeing the *output* from the child process with the prompt, and then fail the `agc -v account activate`.

Maybe the subprocess output processing needs to be de-buffered? Anybody who actually knows Go want to chime in?

adamnovak:
I think I'm going to take a different approach here, actually. Instead of forwarding along all the characters of standard input, I'm going to try and detect when the CDK throws up an MFA prompt, prompt the user in a blocking way for a line of text, and then forward along that line of text.

adamnovak:
OK, this now works for me; I can make it through the AGC [tutorial](https://aws.github.io/amazon-genomics-cli/docs/getting-started/helloworld/) and run a workflow, and at each step I can authenticate with MFA to both AGC and the CDK.

Typing the MFA code is a bit of a pain when not in `-v` mode, since the progress spinner/throbber is going to print itself over whatever you are trying to type, but I've managed to set it up so that at least the prompt is visible, and the spinner will display that it is waiting for an MFA cod to be entered.

I also had to touch some of the bundling code to get the system working for me, since Docker wasn't able to mount into the 700-permissioned temp directories it was using on my machine.

If I wait long enough between steps I hit #254 and our bot deletes buckets and instances due to their lacking an Owner tag, but fixing that is going to be a separate change (and we may just teach the bot about `agc-user-email`).

samwaln:
Since running `agc account activate` without `-v` is not a great experience, should we document the recommended way to activate AGC with MFA?

adamnovak:
The right answer is probably to turn off echo; the Python `awscli` tool doesn't echo the MFA codes to the terminal, and that's presumably more secure and sidesteps the problem of making the throbber know when input is being collected.

It looks like we'd have to add a dependency on [golang.org/x/crypto/ssh and then call its terminal.ReadPassword to do it that way.](https://pkg.go.dev/golang.org/x/crypto/ssh/terminal?utm_source=godoc#ReadPassword).

samwaln (dismissed):

markjschreiber:
The build of this PR seems to be failing on an expected test result that is associated with another of your PRs. You may need to merge this or make the appropriate test changes.

```
     --- FAIL: TestWorkflowRunTestSuite/TestRunWorkflow_UploadToS3Failed (0.00s)
        workflow_run_test.go:381: 
            	Error Trace:	workflow_run_test.go:381
            	Error:      	Error message not equal:
            	            	expected: "unable to run workflow: unable to upload s3://TestOutputBucket/project/TestProject1/userid/bender123/context/TestContext1/workflow/TestLocalWorkflowName1/workflow.zip: cannot upload to S3"
            	            	actual  : "unable to run workflow: cannot upload to S3"
            	Test:       	TestWorkflowRunTestSuite/TestRunWorkflow_UploadToS3Failed
```

adamnovak:
It looks like that PR got merged, and I've merged it back in here. The tests pass for me locally.

packages/cli/internal/pkg/aws/cdk/execute_cdk.go
```diff
@@ -82,12 +93,59 @@ func processOutputs(stdout *bufio.Scanner, stderr *bufio.Scanner, executionName
 	}
 	go func() {
 		defer wait.Done()
+		// We can't just scan through lines, because the MFA prompt is a
+		// partial line and we need to see it. So scan runes instead and do our
+		// own line buffering.
```
nbraid, line 98: Are there tests we can write/update for this behavior?

adamnovak: There are no existing tests for `processOutputs` that I can update.

There are tests [like this](https://github.com/aws/amazon-genomics-cli/blob/a5338edad6e9b4cc9eebc0ed938e956cce15a17b/packages/cli/internal/pkg/aws/cdk/execute_cdk_test.go#L84-L97) for `executeCdkCommandAndCleanupDirectory`. It's a little hard for me to put together why they're doing any of the things they're doing; they don't appear to be documented.

I think what's happening is the tests use fake CDK child process, which is actually [this function](https://github.com/aws/amazon-genomics-cli/blob/a5338edad6e9b4cc9eebc0ed938e956cce15a17b/packages/cli/internal/pkg/aws/cdk/execute_cdk_test.go#L133), and tell it to do one of a few things by passing it some particular arguments. Then they check to see if e.g. a failing exit code was properly detected, or if a line of output was properly parsed.

To use this approach to test the MFA prompt detection and parsing, and the proper forwarding of user input, we would have to:
1. Extend the child process to have a mode where it prints the MFA prompt and tries to read an MFA code. Easy.
2. Get the child process to make sure that the *right* MFA code was received, and somehow get the parent test to fail if it wasn't. I *think* we could just make the child process exit with failure and have that happen. This is doable.
3. Disconnect `processOutputs`, inside the `executeCdkCommandAndCleanupDirectory` function under test, from the real standard input and output streams of the process running the test, and connect it instead to some kind of streams that the test can manipulate, so that the test can watch for the forwarded MFA prompt and provide the fake MFA code. I *think* that can translate to something like "mock `fmt.Printf` and `fmt.Scanln`", if we're comfortable *requiring* that those *particular* functions be used as part of the contract of the function we're testing (or if we're just willing to change the test if we decide to use something else later). But I don't really have much of an idea of how to go about doing this.

It looks like the mocking stuff in use here already, for `os` for checking if the temporary directory removal method is called as expected, is actually calling [some generated code that lives in this repo](https://github.com/aws/amazon-genomics-cli/blob/29f8e9ee012e09ca294547ea3dba225304b36bfa/packages/cli/internal/pkg/mocks/io/mock_interfaces.go). There's [documentation here](https://github.com/aws/amazon-genomics-cli/blob/3d29cae963cd1a0afbbd6cddad42860b693a4c10/packages/cli/TESTING.md#mocks) about how this mock code is to be generated, from an "interface" file.

Actually, it looks like the [interface file](https://github.com/aws/amazon-genomics-cli/blob/29f8e9ee012e09ca294547ea3dba225304b36bfa/packages/cli/internal/pkg/mocks/io/interfaces.go) isn't just for `os` but actually covers a bunch of different system interfaces. Each of them has to be created and attached to the "controller" [like this](https://github.com/aws/amazon-genomics-cli/blob/a5338edad6e9b4cc9eebc0ed938e956cce15a17b/packages/cli/internal/pkg/aws/cdk/execute_cdk_test.go#L48-L49), presumably.

So I could add a `Fmt` type to that interface file, and rerun the mock generator to get a `MockFmt`, which I could then import and attach to the tests' controller. Then I could work out exactly what `fmt.Printf` and `fmt.Scanln` calls I expect to see, and what I would need them to return, and I could program the mock to expect and return those values for that test.

But I'm not entirely sure that's going to work. For no documented reason, the actual code is [using package-scope variables `osRemoveAll` and `execCommand`](https://github.com/aws/amazon-genomics-cli/blob/304a1787fe928e0b5832e535657594f73ad510f5/packages/cli/internal/pkg/aws/cdk/execute_cdk.go#L18-L21) instead of `os.RemoveAll` and `os/exec.Command`. The tests are, also for no documented reason, [swapping these out](https://github.com/aws/amazon-genomics-cli/blob/a5338edad6e9b4cc9eebc0ed938e956cce15a17b/packages/cli/internal/pkg/aws/cdk/execute_cdk_test.go#L50-L54). `osRemoveAll` is in particular swapped fro something from the `MockOS`.

Assuming that the tests are doing this for a good reason, that would mean that parts of the Go standard library that live on global objects, like `os.RemoveAll` and also `fmt.Printf`, cannot actually be mocked in the normal way and replaced out from under the code under test; the code has to be adjusted to cooperate and to allow test functions to be swapped in.

So that in turn would mean I can't really test this as currently written, using the mocking approach that the project usually uses; I'd have to go in and add a `fmtPrintf` and `fmtScanln` that could be swapped out by the test code.

And at that point, really I'd want to just redesign this a bit for better testability: abstract out the thing that is used to prompt for and receive the MFA code, and use a console-based one when running for real, and a special test one when testing.

What do you think is the best approach here, @nbraid? Am I right in my guesses about why the existing test code is doing what it is doing?

nbraid: I'll admit to being a Go novice, and as confused by a lot of what you called out as you are 🙂

If you have refactoring that you think ought to be done, then feel free.

I think what I mostly want to ensure is that changing from simply logging the lines to manually splitting runes and looking for newlines is working, as well as accepting some MFA input.

I think the existing tests somewhat cover the former, although none of the [test commands](https://github.com/aws/amazon-genomics-cli/blob/a5338edad6e9b4cc9eebc0ed938e956cce15a17b/packages/cli/internal/pkg/aws/cdk/execute_cdk_test.go#L155-L158) cover multiple lines at the moment.

For the latter, I think adding a test command that prints the MFA prompt like you suggested, and then verifying that we wait for and pass along input before proceeding would suffice?

Maybe this is harder than I'm realizing. I assume it's made more complex by the fact that we almost entirely capture stdout/stderr of the CDK command and end up piping it through Zerolog rather than using the Go primitives? It doesn't look like the code has any mechanism to pass in pipes to get access to the input/output of the command itself. Would something like that work? Maybe that's what you're referring to by "use a console-based one when running for real, and a special test one when testing."

Would something like

```go
func invokeCdk(commandArgs []string) {
	invokeCdkWithPipes(commandArgs,
		debugLogWriter(),
		progressBarWriter(),
		os.Stdin,
	)
}

func invokeCdkWithPipes(commandArgs []string, stdout io.Writer, stderr io.Writer, stdin io.Reader){
	cmd := execCommand("cdk", commandArgs...)
	cmd.Stdout = stdout
	cmd.Stderr = stderr
	cmd.Stdin = stdin
}
```

allow you to get access to the inputs/outputs from your test and verify them as you need to?

adamnovak: I think the tests need to talk to the I/O of the main Go process, rather than the (fake) CDK process. I can probably swap out a couple of streams, I think; I don't want you to have to pass in streams to get MFA on to all the functions.

packages/cli/internal/pkg/aws/cdk/deploy_app.go
```diff
@@ -11,6 +11,15 @@ var mkDirTemp = os.MkdirTemp
 
 func (client Client) DeployApp(appDir string, context []string, executionName string) (ProgressStream, error) {
 	tmpDir, _ := mkDirTemp(appDir, "cdk-output")
+	// The temp directory defaults to 700 permissions, which can prevent Docker
+	// from mounting directories under it, which can prevent the CDK's bundling
+	// process from working, with errors like:
+	// docker: Error response from daemon: error while creating mount source path '...': mkdir ...: permission denied.
+	// So we need to change the permissions
```
adamnovak, line 18: @nbraid Has the Docker build process I am talking about here been removed? This is a bit unrelated to the point of this PR, and I'd like to remove it, but I needed it when I added it for running the CDK to actually work (along with needing the MFA implementation).

adamnovak:
@nbraid I've added tests for the MFA code prompting and CDK output parsing logic; I think these ought to be enough.

packages/cli/internal/pkg/aws/cdk/deploy_app.go
elliot-smith: Docker has been removed in the current build

nbraid (approved):

adamnovak:
@nbraid It looks like because a conflict was created before this was merged, it needs to be re-reviewed. Can you organize getting two people (maybe you and @samwaln) together to review and merge this before another conflict is created?

elliot-smith (approved):

markjschreiber (approved):

Issue #185, opened 2021-11-18
Remove requirement for Docker during Context Deploy


markjschreiber:
[//]: #  (Before raising the feature request, please check to see if an existing feature request already exists.)

**Description**

[//]: #  (A short description of the feature you are proposing.)
With the addition of WES as a Lambda service we started deploying the required python libs with containerization (the cdk default). This adds two minutes to each deployment and seems to occasionally get stuck during the container build. Adds a dependency on Docker.

**Use Case**

[//]: #  (Why do you need this feature?)

We need to reduce context deployment time and make it as reliable as possible.

**Proposed Solution**

[//]: #  (Please include prototype/workaround/sketch/reference implementation.)

Look into providing the required python libs as a layer/ zip file from our public assets account or add to customers S3 during account initialization?

**Other information**

[//]: #  (detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. associated pull-request, stackoverflow, slack, etc)

eddie1183:
Merged and will be part of the next release.

PR #234, opened 2021-12-17
chore: Use CDK v2

### LINKED ISSUE #87: Add support for S3 server-side-encrypted-KMS (SSE-KMS)
[//]: #  (Before raising the feature request, please check to see if an existing feature request already exists.)

**Description**

AGC cannot be used if server-side-encryption with KMS is being used in the aws account.

Steps to reproduce:

* perform cdk bootstrap -> `BucketA` (generates cdk s3 bootstrap bucket)
* create your data bucket: e.g. `s3://hidden95` -> `BucketB`
* add bucket policy enforcing kms encryption on upload of content. AWS recommended way to enforce SSE KMS encryption on upload
```
{
    "Version": "2012-10-17",
    "Id": "PutObjPolicy",
    "Statement": [
        {
            "Sid": "DenyIncorrectEncryptionHeader",
            "Effect": "Deny",
            "Principal": "*",
            "Action": "s3:PutObject",
            "Resource": "arn:aws:s3:::hidden95/*",
            "Condition": {
                "StringNotEquals": {
                    "s3:x-amz-server-side-encryption": "aws:kms"
                }
            }
        },
        {
            "Sid": "DenyUnEncryptedObjectUploads",
            "Effect": "Deny",
            "Principal": "*",
            "Action": "s3:PutObject",
            "Resource": "arn:aws:s3:::hidden95/*",
            "Condition": {
                "Null": {
                    "s3:x-amz-server-side-encryption": "true"
                }
            }
        }
    ]
}
```
* `agc account activate --bucket hidden95`
* `agc context deploy --context spotContext`

Result: agc fails, since it cannot upload to the S3 bucket without adding sse-kms to the upload.

**Use Case**

Large companies often use S3 bucket policies to prevent upload of unencrypted objects to S3, as suggested in [amazon recommend sse-kms](https://aws.amazon.com/blogs/security/how-to-prevent-uploads-of-unencrypted-objects-to-amazon-s3/)

This means that AGC in its current state cannot be used by these companies.

**Proposed Solution**

Suggested by @elliot-smith 
To enable SSE-KMS, AGC will need the following modifications:

* The AGC command to upload S3 assets will need to optionally supply the required "x-amz-server-side-encryption" header
* The Batch instances will need to have the "kms:Decrypt" IAM property so that the key can be decrypted before it is downloaded
* The documentation will need to be updated to specify this and most likely call out that enabling this functionally requires additional minimum IAM policies

**Other information**

[//]: #  (detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. associated pull-request, stackoverflow, slack, etc)

tneely:
**Description of Changes**

Upgrading AGC to use the latest version of CDK v2. This may additionally fix #87, as this version includes https://github.com/aws/aws-cdk/pull/17668, but it requires more investigation.

Deployed and destroyed both cromwell and miniwdl contexts. Ran `hello` workflow for each successfully.

**Checklist**

----

### DIFF
--- a/packages/cdk/apps/context/app.ts
+++ b/packages/cdk/apps/context/app.ts
@@ -1,5 +1,5 @@
 #!/usr/bin/env node
-import { App } from "monocdk";
+import { App } from "aws-cdk-lib";
 import "source-map-support/register";
 import { APP_NAME, APP_TAG_KEY, CONTEXT_TAG_KEY, PRODUCT_NAME, PROJECT_TAG_KEY, USER_EMAIL_TAG_KEY, USER_ID_TAG_KEY } from "../../lib/constants";
 import { ContextAppParameters } from "../../lib/env";
--- a/packages/cdk/apps/core/app.ts
+++ b/packages/cdk/apps/core/app.ts
@@ -1,5 +1,5 @@
 #!/usr/bin/env node
-import { App } from "monocdk";
+import { App } from "aws-cdk-lib";
 import "source-map-support/register";
 import { getContextOrDefault } from "../../lib/util";
 import { CoreStack } from "../../lib/stacks";
--- a/packages/cdk/lib/common/BucketOperations.ts
+++ b/packages/cdk/lib/common/BucketOperations.ts
@@ -1,13 +1,14 @@
-import { Arn, Construct } from "monocdk";
-import { Bucket, IBucket } from "monocdk/aws-s3";
-import { IRole } from "monocdk/aws-iam";
+import { Arn, ArnFormat } from "aws-cdk-lib";
+import { Bucket, IBucket } from "aws-cdk-lib/aws-s3";
+import { IRole } from "aws-cdk-lib/aws-iam";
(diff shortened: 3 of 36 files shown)


elliot-smith (approved):

IllyaYalovyy (approved):

adamnovak:
I think that this change has cause my attempts to deploy contexts to fail with a complaint from Docker:

```
"--platform" is only supported on a Docker daemon with experimental features enabled
```

This didn't happen with CDK1. I think CDK2 passes `--platform` to Docker calls all the time, even when not trying to cross-build.

There's #185 about removing Docker as a requirement for deploying contexts, but nothing about setting up Docker, or any particular version or configuration of Docker, is listed in the one-time setup section in the README.

What Docker version and configuration is necessary to use AGC after this change?

adamnovak:
It looks like full support for `--platform` comes with Docker API 1.38 or later, and the machine I'm working on has a Docker that only supports up to 1.37.

PR #240, opened 2021-12-23
feat: Enable hosting assets in AWS commercial regions

a-li:
Issue #, if available:

**Description of Changes**
We publish ECR artifacts to `us-west-2` and replicate in `us-east-1`. By default, AGC always pulls images from `us-east-1`. However, some customers in global regions outside of US have restrictions on pulling containers from outside regions (e.g. us-east-1). 

As part of this change, we will replicate the assets to every AWS commercial region and pull from whatever region the customer runs AGC in. 

Tested locally with different AWS regions and it's able to fetch the assets from the correct location/region.

**Checklist**

----

### DIFF
--- a/packages/cli/internal/pkg/cli/context/manager.go
+++ b/packages/cli/internal/pkg/cli/context/manager.go
@@ -65,6 +65,7 @@ type Manager struct {
 	Ssm       ssm.Interface
 	ecrClient ecr.Interface
 	imageRefs map[string]ecr.ImageReference
+	region    string
 
 	baseProps
 	contextProps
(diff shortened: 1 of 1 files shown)


packages/cli/internal/pkg/cli/context/manager.go
```diff
@@ -120,6 +120,7 @@ func NewManager(profile string) *Manager {
 		baseProps: baseProps{homeDir: homeDir},
 		imageRefs: environment.CommonImages,
 		ecrClient: aws.EcrClient(profile),
+		region:    aws.Region(profile),
```
elliot-smith, line 123: Just to double check, when the profile is default (i.e. not passed through), does the region used the one under default/default to us-east-1 if its not set in ~/.aws/profile

elliot-smith (approved):

packages/cli/internal/pkg/cli/context/manager.go
a-li: Yes, it takes the default region if none is provided. 
aws.Region calls getProfileConfig which fetches for the default.

* https://github.com/aws/amazon-genomics-cli/blob/a1ba41f9085af37fe2f69b65a26ac846ddf92c27/packages/cli/internal/pkg/aws/provider.go#L134-L138
* https://github.com/aws/amazon-genomics-cli/blob/a1ba41f9085af37fe2f69b65a26ac846ddf92c27/packages/cli/internal/pkg/aws/provider.go#L146-L156

abassyiouni (approved):

PR #262, opened 2022-01-10
fix: nest batch artifacts and disable bucket clean up

tneely:
**Description of Changes**

The construct we use to upload batch artifacts prunes the target bucket by default. I have disabled that functionality and also moved the artifacts one level deeper in the parent bucket. This will keep the root folder clean and allow us to reenable pruning if we ever need since it won't target the whole bucket.

`agc account activate` and ran hello workflows for cromwell and miniwdl contexts to verify they still run successfully.

**Checklist**

----

### DIFF
--- a/packages/cdk/apps/core/app.ts
+++ b/packages/cdk/apps/core/app.ts
@@ -37,7 +37,7 @@ new CoreStack(app, `${PRODUCT_NAME}-Core`, {
     },
     {
       name: "installed-artifacts/s3-root-url",
-      value: `s3://${bucketName}/batch-artifacts`,
+      value: `s3://${bucketName}/artifacts/batch-artifacts`,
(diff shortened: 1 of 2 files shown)


a-li (approved):

nbraid (approved):

packages/cdk/apps/core/app.ts
```diff
@@ -37,7 +37,7 @@ new CoreStack(app, `${PRODUCT_NAME}-Core`, {
     },
     {
       name: "installed-artifacts/s3-root-url",
-      value: `s3://${bucketName}/batch-artifacts`,
+      value: `s3://${bucketName}/artifacts/batch-artifacts`,
```
wleepang, line 40: How about make this "_common" or "_agc" in case we want to put other things under this prefix?

Issue #211, opened 2021-12-07
`exec user process caused: exec format error` with `agc logs engine`


multimeric:
[//]: #  (Please do not include your AWS account or credentials as part of the bug request.)

**Describe the Bug**

I'm getting `standard_init_linux.go:228: exec user process caused: exec format error`

**Steps to Reproduce**

* Run a workflow using  `agc workflow run some_flow --context some_context`
* `agc logs engine --context some_context`

**Relevant Logs**

```
2021-12-07T14:06:06+11:00 𝒊  Showing engine logs for 'T2Context'
Tue, 07 Dec 2021 13:57:40 +1100 standard_init_linux.go:228: exec user process caused: exec format error
```

**Expected Behavior**

The logs to be shown.

**Actual Behavior**

The logs are not shown; instead I get this error.

Operating System:
AGC Version: 1.1.2
Was AGC setup with a custom bucket: No
Was AGC setup with a custom VPC: No

multimeric:
Okay I have some more info. I looked in the actual cloudwatch log group for this (`NextflowEngineEngineLogGroup`), and this exact message appeared, which means that it's an actual error thrown by the engine and not by the `agc` as I mistakenly thought.

multimeric:
Okay I've found another clue. When I look at the actual job definition of the job that's failing, it seems that the command is not correct, because it tries to run the actual zip file. This would explain why golang (ie Docker) is claiming this is an invalid executable. I have again replaced sensitive information with ALL_CAPS placeholders:
```bash
aws batch describe-jobs --jobs MY_JOB_ARN --query jobs[0].container.command 
[
    "s3://agc-MY-ACCOUNT-ID-MY-REGION-1/project/MY-PROJECT/userid/michaelrmi2UTCAG/context/MY-CONTEXT/workflow/MY-WORKFLOW/workflow.zip"
]
```

multimeric:
I wonder if it relates to this: https://github.com/aws/amazon-genomics-cli/blob/e4ad17efa9a3724c7dd9bc2f37430c21ebcf9d4e/packages/cdk/lib/wes_adapter/amazon_genomics/wes/adapters/NextflowWESAdapter.py#L50-L52

multimeric:
Okay it turns out I went in the wrong direction. I was able to eliminate some issues by running the `nf-core` examples, which work fine:

* The `zip` file is indeed the correct `cmd` for the jobs
[...]

PR #263, opened 2022-01-11
fix: show warning when user choose ARM based instance type in verbose mode during context deployment

evanyui:
fix: show warning when user choose ARM based instance type in verbose mode during context deployment

Issue [#211 ] (https://github.com/aws/amazon-genomics-cli/issues/211):

**Description of Changes**
Add logic to show warning message when ARM based instance is defined in project yaml file

Running `agc context deploy -c <context> --verbose` with c6g instance type (ARM based) to verify warning message is shown.

**Checklist**

- [] If this change would make any existing documentation invalid, I have included those updates within this PR

----


packages/cdk/lib/util/instance-types.ts
```diff
@@ -176,6 +176,10 @@ const unLaunchedInstanceTypesByRegion: { [key in string]: { [key in string]: boo
 
 export const getInstanceTypesForBatch = (instanceTypes: InstanceType[] | undefined, computeType: ComputeResourceType, region?: string): InstanceType[] => {
   if (instanceTypes && instanceTypes.length > 0) {
+    const armBased = instanceTypes.filter((instanceType) => isInstanceTypeArmBased(instanceType.toString()));
+    if (armBased) {
+      console.warn("Using ARM based instance type");
```
a-li, line 181: 1. nit: Better log message. (E.g. "ARM based instance type is not supported in Amazon Genomics CLI...")
2. Maybe see if this is a common issue with all ARM-based instances. If that's really the case, I think we should fail the process immediately instead of simply printing a warning message. According to #211, this cause problems down the road which might be confusing to customers if we let me deploy with ARM instances.

tneely: Do we have any data for what engines/workflows actually run on ARM? I'd rather we fail early here if we know it doesn't work rather than warn them and still deploy all the resources only for it to fail later

packages/cdk/lib/util/instance-types.ts
```diff
@@ -190,3 +194,11 @@ const isInstanceTypeSupported = (instanceType: string, region?: string): boolean
   }
   return true;
 };
+
+const isInstanceTypeArmBased = (instanceType: string): boolean => {
+  const instanceTypeFamily = instanceType.split(".")[0];
```
tneely, line 199: You should use CDK's logic for this since it's more likely to handle all current and feature ARM architectures better

https://docs.aws.amazon.com/cdk/api/v1/docs/@aws-cdk_aws-ec2.InstanceType.html#architecture

packages/cdk/lib/util/instance-types.ts
```diff
@@ -190,3 +194,11 @@ const isInstanceTypeSupported = (instanceType: string, region?: string): boolean
...
+  if (instanceTypeFamily.includes("g")) {
```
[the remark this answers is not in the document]
evanyui: Using cdk enum to check instance processor

packages/cdk/lib/util/instance-types.ts
evanyui: Thanks. Will do.

evanyui:
Closing this PR and created a new one https://github.com/aws/amazon-genomics-cli/pull/264 as branch is corrupted

PR #280, opened 2022-01-26
fix: create wes adapter zip on agc release

evanyui:
Issue #, if available: N/A

**Description of Changes**
Remove docker as dependency, now agc will release bundle including wes adapter as a zip and get uploaded during account activation step. Context deploy will then point lambda source to the S3 bucket to create wes adapter lambda.

I was unable to run CodeBuild locally to test the bundle of zip file during release in the pipeline. But I was able to validate by running package-release script locally, install agc from distribution, run account activate and context deploy and then finally run workflow.
[Updated]
I have just pulled this PR and ran CodeBuild. The zip file appears to be present in .agc under wes/ after running .install.sh script. 

**Checklist**

----


.gitignore
```diff
@@ -10,3 +10,6 @@ vendor
 /docs/
 /site/content/en/docs/Reference/agc*
 /dist/
+*.zip
+/packages/cdk/lib/wes_adapter/package
+/packages/cdk/lib/wes_adapter/wes_adapter
```
IllyaYalovyy, line 15: Can these files redirected to a sub-folder? e.g. "/packages/cdk/lib/wes_adapter/dist/..."

packages/cdk/Makefile
```diff
@@ -15,4 +15,6 @@ init:
 	npm ci --silent
 
 release:
+	cd ./lib/wes_adapter/ && ./install.sh
+	cd ../..
```
IllyaYalovyy, line 19: I think there is a better (more reliable way) of doing this:
```
( cd ./lib/wes_adapter/ && ./install.sh )
```

packages/cdk/lib/stacks/core-stack.ts
```diff
@@ -7,6 +7,7 @@ import { Construct } from "constructs";
 import { PRODUCT_NAME, APP_NAME, VPC_PARAMETER_NAME } from "../constants";
 import { BucketDeployment, Source } from "aws-cdk-lib/aws-s3-deployment";
 import * as path from "path";
+import * as os from "os";
```
IllyaYalovyy, line 10: how is it different from doing `import os from 'os';`?

packages/cdk/lib/util/index.ts
```diff
@@ -124,9 +125,13 @@ export const renderPythonLambda = (
   codePath: string,
   environment: Record<string, string>
 ): PythonFunction => {
-  return new PythonFunction(scope, id, {
+  const account: string = process.env.CDK_DEFAULT_ACCOUNT!;
+  const region: string = process.env.CDK_DEFAULT_REGION!;
+  const bucketName = `${APP_NAME}-${account}-${region}`;
```
IllyaYalovyy, line 130: I think we are already doing this in other places. Can we re-use?

IllyaYalovyy (commented):
The build failed. Please make sure it is fixed in order to unblock merge.

packages/cdk/lib/wes_adapter/install.sh
```diff
@@ -0,0 +1,8 @@
+#! /bin/bash
+
+rm -rf ./package && mkdir ./package
+pip3 install -r requirements.txt --target ./package && cd ./package && zip -r ../wes_adapter.zip .
+cd ..
```
tneely, line 5: I'm not a huge fan of jumping around directories like this. It's hard to track and can lead to weird bugs if a developer forgets what the current working dir is.

Consider wrapping the call in parentheses to avoid changing the working dir: `(cd ./package && zip -r ../wes_adapter.zip .)`

packages/cdk/lib/wes_adapter/install.sh
```diff
@@ -0,0 +1,8 @@
...
+zip -gr ./wes_adapter.zip ./rest_api
+zip -gr ./wes_adapter.zip ./amazon_genomics
+zip -g ./wes_adapter.zip ./index.py
```
tneely, line 8: You should be able to reduce this to a single line: `zip -gr ./wes_adapter.zip ./rest_api ./amazon_genomics ./index.py`

tneely (changes requested):
I understand we're trying to get this out with the next release, but I'd like to see a fast follow here to move the WES adapter code into its own package. Now that we're zipping it separately from CDK it deserves its own package and Makefile with install, test, and release commands

.gitignore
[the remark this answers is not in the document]
IllyaYalovyy: I think we should move it to more appropriate place.

packages/cdk/lib/util/index.ts
```diff
@@ -124,9 +125,13 @@ export const renderPythonLambda = (
...
+  return new Function(scope, id, {
     vpc,
-    entry: codePath,
+    code: Code.fromBucket(Bucket.fromBucketName(scope, "WesAdapter", bucketName), "wes/wes_adapter.zip"),
```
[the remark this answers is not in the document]
IllyaYalovyy: +1 for option 1)

.gitignore
evanyui: the 'package' sub is supposed to be the dist folder. I can move wes_adapter.zip back in there to keep all artifacts in the same place.

packages/cdk/Makefile
evanyui: Done, thanks!

packages/cdk/lib/wes_adapter/install.sh
evanyui: Done, Thanks

packages/cdk/lib/wes_adapter/install.sh
evanyui: Done, Thanks

packages/cdk/lib/util/index.ts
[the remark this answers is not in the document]
evanyui: Moved function to engine construct

packages/cdk/lib/util/index.ts
```diff
@@ -124,9 +125,13 @@ export const renderPythonLambda = (
   codePath: string,
```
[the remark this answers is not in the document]
evanyui: Removed. Thanks

packages/cdk/lib/stacks/core-stack.ts
```diff
@@ -80,6 +81,13 @@ export class CoreStack extends Stack {
       },
     });
 
+    new BucketDeployment(this, "WesAdapter", {
+      sources: [Source.asset(path.join(os.homedir(), ".agc", "wes"))],
+      destinationBucket: this.bucket,
+      destinationKeyPrefix: "wes",
+      prune: true,
```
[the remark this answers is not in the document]
evanyui: I tested it by running account activation a few times, (having the same thought that it would delete the files in the whole bucker), but it didn't. I am guessing this one deploys to a sub dir inside the bucker so it only cleans the .zip file which is our intention. But I will run a few more times today after I address all the other comments.

packages/cdk/Makefile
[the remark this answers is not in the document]
evanyui: Done thanks!

packages/cdk/lib/util/index.ts
evanyui: Although we are doing the same thing in the app stack, our code from go calls the cdk stacks separately (current change is in core stack)

packages/cdk/lib/stacks/core-stack.ts
evanyui: IDE shows there is no default export from 'os', which I thought they had, maybe this node version is different I am not sure. But I have now change to explicitly import only 'homedir'

.gitignore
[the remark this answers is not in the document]
evanyui: Moved all distributions to a dist folder including the zip and added a .gitignore for wes_adapter package.

evanyui:
> I understand we're trying to get this out with the next release, but I'd like to see a fast follow here to move the WES adapter code into its own package. Now that we're zipping it separately from CDK it deserves its own package and Makefile with install, test, and release commands

Now wes_adapter has its own package, makefile.
Although work is still required to add tests into make test step as a future task. 
For this commit, I would leave it out first since there were some complications finding the test dependencies locally and due to time constraints.
However, changes are tested in the manner to running make release, install, and ran a few workflows successfully. 
S3 bucket confirmed to not been deleted by the prune flag because of the prefix (it cleans only subdirectory)

packages/cdk/lib/stacks/core-stack.ts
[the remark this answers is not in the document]
evanyui: Tested again, prune seems to only clean up the subdirectory as I have a destinationKeyPrefix here.

tneely (dismissed):

IllyaYalovyy (dismissed):

tneely (approved):

IllyaYalovyy (approved):

Issue #269, opened 2022-01-18
Ongoing cost reduction options


multimeric:
**Description**

I would like the AGC to offer greater control over AWS resources that have an ongoing cost associated with them. Particularly the NAT Gateway and VPC Endpoints.

[//]: #  (A short description of the feature you are proposing.)

**Use Case**

Many AWS users including myself are concerned about keeping costs down. In particular my use case involves keeping the AGC stack "up" for an extended period of time to support a bioinformatics web tool, so ongoing costs are a big problem.

In particular, the main ongoing costs I would like to see addressed are:
* The NAT Gateway. My monthly bill from NAT from the AGC was $41.72 USD
* VPC Endpoints. My month bill from these endpoints was $21.24 USD

Of course, these costs are dwarfed by the EC2 compute costs and data transfer costs in many use cases, but for my above scenario they are not, which is why I would like the AGC to provide the ability to help address these.

**Proposed Solution**

I don't understand the AGC stack fully, but I suspect that both of these services are only specifically needed by the "head job", ie the workflow manager job. I can't imagine that average bioinformatics jobs require internet access and/or AWS access. So a simple solution then is to put the head node into a public subnet, with a heavily locked-down security group, so that it can receive no public traffic at all. Then, I believe we could drop both the VPC Endpoints and the NAT gateway used by that job, since the public subnet allows the same access.

It might be possible to modify the default stack to do this, without breaking any use cases. However if this is not possible, I could envisage a new key in the context object in the `agc-project.yaml`, along the lines of `publicSubnet: true`, which would enable this functionality.

markjschreiber:
Hi, thanks for your feedback. 

Most nodes require some access to the internet or AWS services like S3, Parameter store, ECR and CloudWatch. For example, workflow tasks are containerized and frequently access inputs from S3 and/ or shared storage like EFS. They also need to communicate with the Batch cluster and CloudWatch to coordinate and provide logging.

The current configuration, using private subnets and VPC endpoints is the best practice for security posture and the use of private subnets and endpoints is a requirement of many organizations. As you note however, it does add some on-going costs.

It is possible to use AGC with an existing VPC to avoid any additional costs of having an extra VPC. This VPC will still require private subnets but it could help avoid paying for additional NAT gateways. To do this you would provide the ID of the existing VPC when activating AGC. https://aws.github.io/amazon-genomics-cli/docs/reference/agc_account_activate/ 

[...]

PR #283, opened 2022-01-27
feat: add "public subnet" mode for lower ongoing costs

multimeric:
Issue #, if available: #269

**Description of Changes**

* Add `publicSubnets` option to core stack and batch stack
  * For the core stack, this flag makes the stack exclude the VPC endpoints and NAT gateway
  * For the batch stack, this puts the EC2 instances into the public subnet
  * Together, this will reduce ongoing costs for users
* Take WES Adapter Lambda out of the VPC (see [explanation here](https://github.com/aws/amazon-genomics-cli/pull/283#issuecomment-1031419841))
* Usage is as follows:
    * `agc account activate --publicSubnets` will create the `Agc-Core` stack, but without the endpoints/NAT
    * In the context object in `agc-project.yaml`, there is now a `publicSubnets` key, which defaults to false. If the core stack was created with `--publicSubnets`, this *must* be set to true or the deployment will fail

* Updated the golang tests
* Manually verified:
    * Core stack
        * I ran `aws cloudformation get-template` on stacks generated in both ways
        * By default: [default_core.txt](https://github.com/aws/amazon-genomics-cli/files/8013919/default_core.txt)
        * With `--publicSubnets`: [public_core.txt](https://github.com/aws/amazon-genomics-cli/files/8013921/public_core.txt)
        * Note that the NAT gateways and VPC endpoints are only present in the first stack
    * Context stack
        * I ran `aws cloudformation get-template` on stacks generated both ways
        * By default: [default_context.txt](https://github.com/aws/amazon-genomics-cli/files/7951355/default_context.txt)
        * With `publicSubnets: true` in the config file: [public_context.txt](https://github.com/aws/amazon-genomics-cli/files/7951358/public_context.txt)
        * Note how the public stack sets the compute environment to use public subnets, when you look these subnets up in the `public_core.txt` stack above, whereas the default stack uses private subnets.
    * Successfully ran the `examples/demo-nextflow-project` after deploying a public account + context stack

**Checklist**

- [ ] I have added unit tests that prove my fix is effective or that my feature works. *I have not yet done this, but I'm interested in the recommended way of doing this. Ideally we would write assertions using CDK but I note that there is no existing example of this being done.*
- [x] I have linted my code before raising the PR. *I ran `make format` at least. Is that what you want?*
- [x] Title of this Pull Request follows Conventional Commits standard: https://www.conventionalcommits.org/en/v1.0.0/.

----


multimeric:
Okay, I finally got the chance to test this out with a real workflow, and I realised that, when we remove the NAT gateway, the WES Adapter lambda gets put into an isolated subnet, and therefore cannot communicate with AWS. Further, [you can't put lambdas into a public subnet](https://stackoverflow.com/a/52994841/2148718). However the simple and clean solution was just to take that lambda out of the VPC entirely: it doesn't actually need to be in one, since it's called by the API gateway and all of its actions are performed through the AWS API. Once I made that change (which is now pushed), the workflow ran successfully! I used `examples/demo-nextflow-project`.

I've also pulled some upstream changes from the main branch published since I made this PR.

multimeric:
Hi @markjschreiber is there any chance that this might get reviewed? I am happy to fix the merge conflicts again if a review can be scheduled.

markjschreiber:
> Hi @markjschreiber is there any chance that this might get reviewed? I am happy to fix the merge conflicts again if a review can be scheduled.

Sorry for the delay, I have been working on other bug fixes. I will start reviewing today and also ask another engineer to take a look.

packages/cdk/apps/core/app.ts
```diff
@@ -15,11 +15,13 @@ const agcVersion = getEnvString(app.node, "AGC_VERSION");
 const vpcId = getContextOrDefault<Maybe<string>>(app.node, "VPC_ID");
 const bucketName = getContextOrDefault(app.node, `${APP_ENV_NAME}_BUCKET_NAME`, `${APP_NAME}-${account}-${region}`);
 const createNewBucket = getContextOrDefault(app.node, `CREATE_${APP_ENV_NAME}_BUCKET`, "true").toLowerCase() == "true";
+const publicSubnets = getContextOrDefault(app.node, `${APP_ENV_NAME}_PUBLIC_SUBNETS`, "false").toLowerCase() == "true";
```
markjschreiber, line 18: usePublicSubnets might be a better variable name

packages/cdk/lib/constructs/batch.ts
```diff
@@ -58,6 +58,13 @@ export interface ComputeOptions {
    * @default none
    */
   resourceTags?: { [p: string]: string };
+
+  /**
+   * If true, put EC2 instances into public subnets instead of private subnets. This should be used in conjunction
+   * with the ENV_PUBLIC_SUBNETS option
+   * This will result in significantly lower ongoing costs when no job is running.
```
markjschreiber, line 65: as no NAT Gateways will be created

packages/cdk/lib/constructs/batch.ts
```diff
@@ -168,12 +175,17 @@ export class Batch extends Construct {
 
   private renderComputeEnvironment(options: ComputeOptions): IComputeEnvironment {
     const computeType = options.computeType || defaultComputeType;
+    const subnets = {
+      // Even if we use public subnets, CDK will assign security groups that don't allow inbound
```
markjschreiber, line 179: This comment is currently accurate but might be misleading because it is only true if the security groups are not changed in future updates to this stack.

packages/cdk/lib/env/context-app-parameters.ts
```diff
@@ -79,6 +79,12 @@ export class ContextAppParameters {
    * The types of EC2 instances that may be launched in the compute environment.
    */
   public readonly instanceTypes?: InstanceType[];
+  /**
+   * If true, put EC2 instances into public subnets instead of private subnets. This should be used in conjunction
+   * with the ENV_PUBLIC_SUBNETS option
+   * This will result in significantly lower ongoing costs when no job is running.
```
markjschreiber, line 85: Same as above
due to NAT gateways ...

Probably also worth documenting the potential security risk if users also modify security groups

packages/cdk/lib/stacks/core-stack.ts
```diff
@@ -55,6 +55,13 @@ export interface CoreStackProps extends StackProps {
    * @default none
    */
   parameters?: ParameterProps[];
+  /**
+   * If true, use public subnets instead of private subnets, and do not create the VPC endpoints and the NAT gateway.
+   * This will result in significantly lower ongoing costs when no job is running.
```
markjschreiber, line 60: Same comment as above regarding why costs are lower and potential security warning

packages/cdk/lib/stacks/core-stack.ts
```diff
@@ -67,7 +74,8 @@ export class CoreStack extends Stack {
   constructor(scope: Construct, id: string, props: CoreStackProps) {
     super(scope, id, props);
 
-    this.vpc = this.renderVpc(props.vpcId);
+    console.error(`Params: ${JSON.stringify(props)}`);
```
markjschreiber, line 77: If this was added for debug logging it can probably be removed

packages/cdk/lib/stacks/core-stack.ts
```diff
@@ -94,47 +102,54 @@ export class CoreStack extends Stack {
     new CfnOutput(this, "TableName", { value: this.table.tableName });
   }
 
-  private renderVpc(vpcId?: string): IVpc {
+  private renderVpc(vpcId?: string, publicSubnets?: boolean): IVpc {
```
markjschreiber, line 105: If this was added for debug logging it can probably be removed

packages/cdk/lib/stacks/engines/batch-construct.ts
```diff
@@ -65,6 +65,7 @@ export class BatchConstruct extends Construct {
       launchTemplateData: LaunchTemplateData.renderLaunchTemplateData(appParams.engineName),
       awsPolicyNames: ["AmazonSSMManagedInstanceCore", "CloudWatchAgentServerPolicy"],
       resourceTags: Stack.of(this).tags.tagValues(),
+      publicSubnets: appParams.publicSubnets,
```
markjschreiber, line 68: as above, usePublicSubnets might be a better name

site/content/en/docs/Concepts/accounts.md
```diff
@@ -92,6 +92,10 @@ as VPC NAT gateways and VPC Endpoints. If you no longer use Amazon Genomics CLI
 wish to remove the S3 bucket along with its objects as well as the [CloudWatch](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/index.html) logs produced by Amazon Genomics CLI. These are retained
 by default so that you can view workflow results and logs even after deactivation.
 
+However, if you with to have this infrastructure remain deployed, you are able to significantly reduce ongoing costs by using `agc account activate --publicSubnets`.
```
markjschreiber, line 95: `s/with/which/`

packages/cdk/lib/stacks/engines/nextflow-engine-construct.ts
```diff
@@ -58,7 +58,6 @@ export class NextflowEngineConstruct extends EngineConstruct {
     engineLogGroup.grant(adapterRole, "logs:StartQuery");
 
     const lambda = this.renderAdapterLambda({
-      vpc: props.vpc,
```
markjschreiber, line 61: Won't the lambda need the `vpc` reference when it is being used with a custom VPC?

multimeric: Well this is the tricky thing. My "public mode" will not work if the lambdas need to be in a VPC: if it's in a public subnet then lambdas don't work, and if a private subnet then we're paying for a NAT gateway so my changes are pointless, and an isolated subnet would mean no access to AWS services. Therefore the lambda has to be in no VPC for my purposes.

Perhaps I could implement a check that says that the public subnet mode is mutually exclusive with the custom VPC? Would that be sufficient?

packages/cdk/lib/env/context-app-parameters.ts
multimeric: I would like to avoid repeating myself where possible. Can I put this discussion in the documentation?

packages/cdk/lib/stacks/engines/nextflow-engine-construct.ts
multimeric: I have made this change in my recent commits.

multimeric:
Okay, I think I've fixed all the issues from that code review. Thanks for that. Here are example stacks:
* [context_stack.json](https://github.com/aws/amazon-genomics-cli/files/8086195/public_context.txt)
* [core_stack.json](https://github.com/aws/amazon-genomics-cli/files/8086196/public_core.txt)

packages/cdk/lib/stacks/core-stack.ts
```diff
@@ -96,47 +106,53 @@ export class CoreStack extends Stack {
...
     const vpc = new Vpc(this, "Vpc", {
-      gatewayEndpoints: {
-        S3Endpoint: { service: GatewayVpcEndpointAwsService.S3 },
-      },
+      gatewayEndpoints: publicSubnets
+        ? {}
+        : {
+            S3Endpoint: { service: GatewayVpcEndpointAwsService.S3 },
+          },
+      subnetConfiguration: publicSubnets ? Vpc.DEFAULT_SUBNETS_NO_NAT : Vpc.DEFAULT_SUBNETS,
```
markjschreiber, line 120: Minor point, `Vpc.DEFAULT_SUBNETS_NO_NAT` will still create one private and one public subnet per AZ. The private subnet will be "isolated" from the internet as there is no NAT. If you want to remove private subnets entirely you could have to provide a customized configuration here

packages/cli/internal/pkg/cli/account_activate.go
```diff
@@ -23,16 +23,17 @@ import (
 const (
 	accountBucketFlag            = "bucket"
 	accountVpcFlag               = "vpc"
-	publicSubnetsFlag            = "publicSubnets"
+	publicSubnetsFlag            = "usePublicSubnets"
```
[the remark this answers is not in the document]
multimeric: Good point. I have done this behaviour inside the CDK stack, but this should be reported earlier when they run the CLI.

packages/cdk/lib/stacks/core-stack.ts
multimeric: Do you think there is any advantage in removing the private subnets? I don't think the stack actually puts anything in them, but I don't think we should expose this as an option since it increases the amount of moving parts in the stack, but I would be happy to just change it to only public subnets.

markjschreiber: I don't think it adds anything related to cost. Can you check if there would be any impact in terms of account level limits on subnets or similar?

multimeric: I agree it won't affect costs. There is a [200-subnet limit on VPCs](https://docs.aws.amazon.com/vpc/latest/userguide/amazon-vpc-limits.html), so I guess this is wasting N subnets, where N is the number of AZs. I guess I might as well change this to only public then.

multimeric: Okay, I've now made it generate only a public subnet.

multimeric:
New updates from the last round of code review. Here are some new example CFN templates. They demonstrate that the VPC only produces public subnets now:

* [public_core.json](https://github.com/aws/amazon-genomics-cli/files/8093896/public_core.txt)
* [public_context.json](https://github.com/aws/amazon-genomics-cli/files/8093894/public_context.txt)

markjschreiber (dismissed):

multimeric:
Please let me know if any more work is needed for this.

markjschreiber:
@multimeric one of the build checks is failing seemingly due to a linting failure (changed import order). You can probably fix this by running `make format` and then commiting the result.

multimeric:
Okay I've pulled from upstream and reformatted the code.

markjschreiber (dismissed):

waliag87 (dismissed):

waliag87:
@multimeric there appears to be a formatting issue again, likely due to my merge from main. If we can resolve that + get it re-reviewed by Mark and I the PR should be ready to merge.

multimeric:
Okay I've fixed the tests again. This problem keeps happening because changes to upstream break the merge, so the faster this can get approved the better.

markjschreiber (approved):

waliag87 (approved):

waliag87:
@multimeric done and merged in.

PR #288, opened 2022-01-31
feat: Prune out dev dependencies from the final CDK bundle

multimeric:
Issue #, if available:

**Description of Changes**

* Run `npm prune --production` after building the CDK bundle
* This removes the dev dependencies, most notably the typescript compiler, from the bundle provided to the user
* This makes the bundle significantly smaller

I built a docker image, pruned the dependencies using `npm prune`, and then ran `agc account activate`. This succeeded without any import errors that might be caused by missing dependencies. It later failed due to missing credentials but I suspect that's unrelated to the CDK installation.

**Checklist**

----

### DIFF
--- a/scripts/cli/install.sh
+++ b/scripts/cli/install.sh
@@ -34,7 +34,7 @@ install_cli () {
 install_cdk () {
     mkdir -p "$BASE_DIR/cdk"
     cp "$SCRIPT_DIR/cdk.tgz" "$BASE_DIR/cdk"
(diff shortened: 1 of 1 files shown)


tneely (approved):

markjschreiber (approved):

multimeric:
Hmm I've just noticed this: https://github.com/aws/amazon-genomics-cli/blob/219596e8edf1666ca46a065bfa1ccc74848180f5/packages/cdk/apps/core/cdk.json#L2

I guess that typescript is not actually a dev dependency since it's being used at runtime for this app? Perhaps it should not be a `devDependency`?

Issue #294, opened 2022-02-01
CDK npm-shrinkwrap.json has become corrupted; claims to be version 2 but is missing bin information


adamnovak:
https://github.com/aws/amazon-genomics-cli/blob/91d6ed42e3237b5b1ee605ffcccc0d2a6d0e63a6/packages/cdk/npm-shrinkwrap.json#L4

The `npm-shrinkwrap.json` file for the CDK code was upgraded to lockfile format version 2, but this upgrade process was botched and `"bin"` entries were not created for the packages that need them.

As noted in https://github.com/npm/cli/issues/4308 this can happen when using `npm@8` on a file generated with `npm@6`, without going through `npm@7`'s migration code.

The resulting `npm-shrinkwrap.json` file results in `npm install` run with NPM 6, 7, or 8 not creating the symlinks in `node_modules/.bin` that allow tools like `prettier` to be executed. The information on what to link there isn't sourced from the packages themselves like with the v1 format, but it's also missing from where it should be stored in the v2 format.

This causes the AGC build to fail thusly:

```
make[2]: Entering directory `/public/groups/cgl/graph-genomes/anovak/build/amazon-genomics-cli/packages/wes_adapter'
./format-code.sh

> cdk@0.1.0 pretest
> eslint --ignore-path .gitignore '{**/*,*}.{ts,tsx}'


> cdk@0.1.0 format
> prettier --write '{**/*,*}.{ts,tsx}'

sh: prettier: command not found
sh: eslint: command not found
make[2]: *** [format] Error 127
```

Reverting to the v1-format npm-shrinkwrap.json in be5c684ca0967baeda7cb8055d24d39b54cd3084 allows the bin symlinks to be created.

Beyond using `.nvmrc`, it may be useful to force a particular NPM version for all developers to prevent problems like this from happening.

adamnovak:
Sorry, this is just a dupe of #292.

Issue #292, opened 2022-02-01
Node executables not being installed on main branch


multimeric:
[//]: #  (Please do not include your AWS account or credentials as part of the bug request.)

**Describe the Bug**

[//]: #  (Please describe the behavior that you are seeing with as many details as possible including logs from AGC, CloudWatch snippets and CloudFormation errors if relevant.)
I'm getting `sh: 1: ts-node: not found` when I `agc account activate` when using the latest code off the main branch.

**Steps to Reproduce**

```
git clone git@github.com:aws/amazon-genomics-cli.git
cd amazon-genomics-cli
make release
./dist/amazon-genomics-cli/install.sh
agc account activate
```

**Relevant Logs**

```
2022-02-01T14:05:56+11:00 𝒊  Activating AGC with bucket '' and VPC ''
Bootstrapping CDK... [-o--] 1.8s                                                                                                                                                                                   2022-02-01T14:06:01+11:00 ✘  sh: 1: ts-node: not found
2022-02-01T14:06:01+11:00 ✘  Subprocess exited with error 127
2022-02-01T14:06:01+11:00 ✘   error="exit status 1"
Error: an error occurred invoking 'account activate'
with variables: {bucketName: vpcId: publicSubnets:false}
caused by: exit status 1
```

**Expected Behavior**

The core stack to be deployed,

**Actual Behavior**

The above error

**Additional Context**

I think this involves the fact that the `npm-shrinkwrap.json` is missing the binaries, when it shouldn't be. If I run `npm shrinkwrap` on the previous (stable) release, I get a shrinkwrap file that includes the binaries, such as the one causing this issue:
```json
    "node_modules/ts-node": {
      "version": "10.4.0",
      "resolved": "https://registry.npmjs.org/ts-node/-/ts-node-10.4.0.tgz",
      "integrity": "sha512-g0FlPvvCXSIO1JDF6S232P5jPYqBkRL9qly81ZgAOSU7rwI0stphCgd2kLiCrU9DjQCrJMWEqcNSjQL02s6d8A==",
      "dev": true,
      "bin": {
        "ts-node": "dist/bin.js",
        "ts-node-cwd": "dist/bin-cwd.js",
        "ts-node-script": "dist/bin-script.js",
        "ts-node-transpile-only": "dist/bin-transpile.js",
        "ts-script": "dist/bin-script-deprecated.js"
      },
    },
```

[...]

PR #299, opened 2022-02-01
fix: force use node v14 in cdk package

evanyui:
Issue #, if available: [294](https://github.com/aws/amazon-genomics-cli/issues/294), [292](https://github.com/aws/amazon-genomics-cli/issues/292)

**Description of Changes**
Force use version 14 when running npm

Ran `make init` which calls `npm ci` with version 15, 16, 17 and verified failure with message engine node version is not v14. Reran `make init` with version 14 succeeded.

v17 logs
```
npm ERR! code EBADENGINE
npm ERR! engine Unsupported engine
npm ERR! engine Not compatible with your version of node/npm: cdk@0.1.0
npm ERR! notsup Not compatible with your version of node/npm: cdk@0.1.0
npm ERR! notsup Required: {"node":"^14","npm":"^6"}
npm ERR! notsup Actual:   {"npm":"8.3.1","node":"v17.4.0"}

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/chuanyut/.npm/_logs/2022-02-01T23_35_58_661Z-debug-0.log
make: *** [init] Error 1
```

v14 logs
```
npm WARN prepare removing existing node_modules/ before installation
added 830 packages in 10.981s
```

**Checklist**

----

### DIFF
--- a/packages/cdk/.npmrc
+++ b/packages/cdk/.npmrc
@@ -1,0 +1,2 @@
+engine-strict=true
+package-lock=false
--- a/packages/cdk/.npmrc
+++ b/packages/cdk/.npmrc
@@ -1,2 +1,1 @@
-engine-strict=true
 lockfile-version=1
--- a/packages/cdk/.npmrc
+++ b/packages/cdk/.npmrc
@@ -1,2 +1,2 @@
 engine-strict=true
-package-lock=false
+lockfile-version=1
--- a/packages/cdk/Makefile
+++ b/packages/cdk/Makefile
@@ -12,7 +12,7 @@ format:
 	npm run format
 
 init: 
-	npm ci
+	npm ci --loglevel error
 
 release:
 	npm shrinkwrap && npm pack --silent | xargs -I '{}' mv {} cdk.tgz
--- a/packages/cdk/Makefile
+++ b/packages/cdk/Makefile
@@ -12,7 +12,7 @@ format:
 	npm run format
 
 init: 
-	npm ci --silent
+	npm ci
 
 release:
 	npm shrinkwrap && npm pack --silent | xargs -I '{}' mv {} cdk.tgz
--- a/packages/cdk/package.json
+++ b/packages/cdk/package.json
@@ -3,10 +3,6 @@
   "version": "0.1.0",
   "main": "lib/index.js",
   "types": "lib/index.d.ts",
-  "engines": {
-    "node": ">=14.0.0",
-    "npm": ">=6.0.0"
-  },
   "scripts": {
     "build": "tsc",
     "format": "prettier --write '{**/*,*}.{ts,tsx}'",
--- a/packages/cdk/package.json
+++ b/packages/cdk/package.json
@@ -4,8 +4,8 @@
   "main": "lib/index.js",
   "types": "lib/index.d.ts",
   "engines": {
-    "node": "^14",
-    "npm": "^6"
+    "node": ">=14.0.0",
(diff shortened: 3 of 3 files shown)


packages/cdk/Makefile
```diff
@@ -12,7 +12,7 @@ format:
 	npm run format
 
 init: 
-	npm ci --silent
```
evanyui, line 15: This is to show failure message when using different versions of node

packages/cdk/.npmrc
```diff
@@ -0,0 +1,2 @@
+engine-strict=true
+package-lock=false
```
evanyui, line 2: Since we are using shrinkwrap.json, I disable producing package-lock.json here

tneely: I don't think this is doing what you think it is. Setting this to false will also ignore the shrinkwrap and effectively bump our dependencies on every install.

a-li (dismissed):

packages/cdk/.npmrc
```diff
@@ -0,0 +1,2 @@
+engine-strict=true
```
tneely, line 1: I think this will cause more problems than it's worth. We don't have strict NPM version requirements at the moment outside of a minimum version of 14. If a package is untested for a new version of NPM that the customer is on, then they are effectively blocked from using AGC on that version.

If you're just looking to prevent accidental lockfile version upgrades, please use [lockfile-version](https://docs.npmjs.com/cli/v8/using-npm/config#lockfile-version) instead.

packages/cdk/Makefile
tneely: I believe `--silent` will still show the final error message. Removing this makes the install a lot noisier

packages/cdk/package.json
```diff
@@ -3,6 +3,10 @@
   "version": "0.1.0",
   "main": "lib/index.js",
   "types": "lib/index.d.ts",
+  "engines": {
+    "node": "^14",
+    "npm": "^6"
+  },
```
tneely, line 9: I don't think we should restrict customers to only v14/v6 of Node/NPM. The lockfile version we use is forwards compatible with all current Node/NPM releases.

If you want to set a minimum version (ie `>=14`), I'm all for it, but this is overkill and restrictive to our users.

tneely (changes requested):

packages/cdk/Makefile
tneely: Hmm I think I'm wrong here actually, though I still don't feel it's necessary to unsilence

evanyui: Done

packages/cdk/.npmrc
evanyui: You are correct.

packages/cdk/.npmrc
evanyui: makes sense, i change rules to strictly allow only version >=14.0.0 (which comes with npm >=6.0.0), then use lockfile-version flag

packages/cdk/package.json
evanyui: Agree. Final revision removes the strict engine rule. Only locking the lockFileversion

packages/cdk/.npmrc
evanyui: Update: removing any engine strict rules and using lockfile-version flag only

tneely (approved):

a-li (approved):

Issue #278, opened 2022-01-25
Reduce initial EBS volume size for miniwdl instances to reduce operational cost


markjschreiber:
[//]: #  (Before raising the feature request, please check to see if an existing feature request already exists.)

**Description**

EC2 batch workers used by miniwdl can use smaller initial disk allocations because most of the scratch space is on EFS in these deployments.

**Use Case**

Reducing disk allocation reduces cost, particularly for larger workflows.

**Proposed Solution**

Investigate the need for a 30 GB disk and a 200 GB disk which are currently used. Determine how much space is actually used in our example workflows and propose a smaller disk size. Update CDK accordingly.

Issue #275, opened 2022-01-21
S3 retry for Cromwell logic doesn't work


elliot-smith:
[//]: #  (Please do not include your AWS account or credentials as part of the bug request.)

**Describe the Bug**

Currently, if an S3 download fails when Cromwell is being setup, the loop logic will fail as the value is being cast to a string. The logic is currently

`
sleep $((7 * "$i"))
`
but it should be 
`
sleep $((7 * i))
`

Also, there is some duplicate logic in there that could be refactored out as well as some additional logs for assisting with troubleshooting

**Steps to Reproduce**

[//]: #  (Include a list of steps that can be used to reproduce the bug.)

Run a workflow that fails to download the S3 contents for the first time.

**Relevant Logs**

`
attempt 1 to copy s3://agc-383881756834-us-west-2/project/Demo/userid/elliotsm3WJrU7/context/myContext/cromwell-execution/Words_With_Vowels/fa3208ab-92b6-445b-9044-35ed9de948a6/call-FindWordsWithLetter/shard-0/script failed
--
/tmp/tmp.pLTmOLrXI/batch-file-temp: line 33: 7 * "1": syntax error: operand expected (error token is ""1"")
grep: /tmp/scratch/agc-383881756834-us-west-2/project/Demo/userid/elliotsm3WJrU7/context/myContext/workflow/words-with-vowels/mieliestronk-words.txt: No such file or directory

`

**Expected Behavior**
The download logic loops

**Actual Behavior**
The download fails and the loop logic is skipped

**Additional Context**

[//]: #  (Add any other context about the problem here. The minimum required information is below)

Operating System:
AGC Version: Built from source on 1/21
Was AGC setup with a custom bucket: No
Was AGC setup with a custom VPC: No

markjschreiber:
Closed by PR #300

PR #300, opened 2022-02-02
feat: don't use ebs auto scale for minidl

markjschreiber:
Issue #, if available:

#278
#275

**Description of Changes**

* Make LaunchTemplate user data aware of workflow orchestrator (engine). Don't use ebs-autoscale when running miniwdl.
* Add code comment  about the need to redeploy contexts when changing the LaunchTemplate.
* Change shutdown to echo error because shutdown won't cause batch to restart a new instance.
* Exposed initial ebs volume size to provision.sh to make it easier to change this value.
* fixed additional typo/ bug from issue 275

Because the changes here modify the ecs-additions scripts a full account deactivate, activate is required.

* deploy miniwdl context and run `demo-wdl-project` workflows. Make sure the autoexpand EBS volume is *not* created on batch workers
* deploy cromwell context and run `demo-wdl-project` workflows. Make sure the autoexpand EBS volume is still created on batch workers

**Checklist**

----

### DIFF
--- a/packages/cdk/lib/artifacts/batch-artifacts/ecs-additions/provision.sh
+++ b/packages/cdk/lib/artifacts/batch-artifacts/ecs-additions/provision.sh
@@ -7,8 +7,7 @@ export OS=`uname -r`
 BASEDIR=`dirname $0`
 
 # Expected environment variables
-#   GWFCORE_NAMESPACE
-#   ARTIFACT_S3_ROOT_URL
+#   ARTIFACT_S3_ROOT_URL (obtained from SSM parameter store)
 #   WORKFLOW_ORCHESTRATOR (OPTIONAL)
 
 printenv
@@ -62,11 +61,15 @@ ARTIFACT_S3_ROOT_URL=$(\
 
 
 # retrieve and install amazon-ebs-autoscale
-cd /opt
-sh $BASEDIR/get-amazon-ebs-autoscale.sh \
-    --install-version dist_release \
-    --artifact-root-url $ARTIFACT_S3_ROOT_URL \
-    --file-system btrfs
+if [ "$WORKFLOW_ORCHESTRATOR" != "miniwdl" ]; then
+  initial_ebs_size=200
+  cd /opt
+  sh $BASEDIR/get-amazon-ebs-autoscale.sh \
+      --install-version dist_release \
+      --artifact-root-url "$ARTIFACT_S3_ROOT_URL" \
+      --file-system btrfs \
+      --initial-size "$initial_ebs_size"
+fi
 
 # common provisioning for all workflow orchestrators
 cd /opt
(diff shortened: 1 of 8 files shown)


packages/cdk/lib/artifacts/batch-artifacts/ecs-additions/provision.sh
```diff
@@ -62,11 +61,15 @@ ARTIFACT_S3_ROOT_URL=$(\
 
 
 # retrieve and install amazon-ebs-autoscale
-cd /opt
-sh $BASEDIR/get-amazon-ebs-autoscale.sh \
-    --install-version dist_release \
-    --artifact-root-url $ARTIFACT_S3_ROOT_URL \
-    --file-system btrfs
+if [ "$WORKFLOW_ORCHESTRATOR" != "miniwdl" ]; then
+  initial_ebs_size=200
```
tneely, line 65: Do you want to use `INITIAL_SIZE` here?

packages/cdk/lib/constants.ts
```diff
@@ -9,102 +11,8 @@ export const USER_EMAIL_TAG_KEY = `${APP_NAME}-user-email`;
...
-- test -f ./ecs-additions/fetch_and_run.sh || shutdown -P now
-- /opt/ecs-additions/provision.sh
+export const ENGINE_CROMWELL = "cromwell";
+export const ENGINE_MINIWDL = "miniwdl";
+export const ENGINE_NEXTFLOW = "nextflow";
 
-- echo "successfully initiated"
---==MYBOUNDARY==--
-`;
+export const wesAdapterSourcePath = path.resolve(path.join(__dirname, "./wes_adapter"));
```
tneely, line 18: This doesn't seem right - isn't it stored outside of the `cdk` package now. I also don't see this being used anywhere

packages/cdk/lib/constructs/launch-template-data.ts
```diff
@@ -0,0 +1,112 @@
...
+              }
+            ]
+          }
+        }
+      }
+    }
+
+runcmd:
+
+# start the amazon-cloudwatch-agent
```
elliot-smith, line 76: Can you please add the change for https://github.com/aws/amazon-genomics-cli/pull/301 here as well?

packages/cdk/lib/artifacts/batch-artifacts/ecs-additions/provision.sh
markjschreiber: That variable is in the `get-amazon-ebs-autoscale.sh` script which is called by this script. I could however made a parameter for this script that can be initialized.

packages/cdk/lib/constants.ts
markjschreiber: Agreed, seems like a left over from my rebase.

packages/cdk/lib/constructs/launch-template-data.ts
markjschreiber: Done

tneely (approved):

elliot-smith (approved):

PR #301, opened 2022-02-02
fix: fix retry logic for ecs-additions

a-li:
Issue #, if available:

**Description of Changes**

Fix ecs-addition retry logics. Currently the retry never happens because after `sleep` execute successful, it never goes into the retry block due to the `||` logic. Changed it to `&&` so it gets correctly executed. 

`make`, `agc account activate` and `agc deploy context`

**Checklist**

----


tneely (dismissed):

packages/cdk/lib/constants.ts
```diff
@@ -100,8 +100,8 @@ runcmd:
 # Setup ecs additions
```
elliot-smith, line 100: The code above also needs to be && instead of pipe
 command -v aws || sleep 5 |

a-li: Good catch. Changed it as well.

tneely (approved):

a-li:
This change has been incorporated as part of #300. Closing this out as the same diff is already merged in main.

PR #306, opened 2022-02-03
fix: Delete bootstrap bucket on account deactivate

tneely:
**Description of Changes**

When deactivating an AGC-CDK bootstrapped account, we do not delete the assets bucket associated with the bootstrap stack, causing subsequent activations to fail.

Activated and deactivated and reactivated account to confirm bucket is deleted and does not block reactivation.

**Checklist**

----

### DIFF
--- a/packages/cli/internal/pkg/aws/s3/delete_bucket.go
+++ b/packages/cli/internal/pkg/aws/s3/delete_bucket.go
@@ -1,0 +1,21 @@
+package s3
+
+import (
+	"context"
+
+	"github.com/aws/amazon-genomics-cli/internal/pkg/cli/clierror/actionableerror"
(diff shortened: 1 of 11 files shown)


elliot-smith:
Can we please update the documentation for this. Account deactivate is now a destructive operation as it will remove all workflow runs, particularly because we call out "Buckets and logs will be preserved."

tneely:
> Can we please update the documentation for this. Account deactivate is now a destructive operation as it will remove all workflow runs, particularly because we call out "Buckets and logs will be preserved."

Buckets and logs are still preserved. This is just deleting the CDK asset bucket which is used to stage things like lambda deployments. There is no valuable customer data that is lost by deleting this bucket.

elliot-smith (approved):

PR #313, opened 2022-02-09
feat: Implement EFS IOPs specification in Project YAML

samwaln:
**Description of Changes**

We want to be able to specify the filesystem used and the throughput. This allows customers with large numbers of large files to accelerate workflows. Filesystem options are S3 and EFS, with a default of S3 for Cromwell and Nextflow and a default of EFS for MiniWDL. Valid throughput values are 1 to 1024.

I have added a test that checks that the default throughput is set. I have run `./scripts/run-dev.sh` and `agc project validate` with a valid agc-project.yaml and an invalid version containing an invalid filesystem.

**Checklist**

----

### DIFF
--- a/packages/cdk/lib/constructs/engines/engine.ts
+++ b/packages/cdk/lib/constructs/engines/engine.ts
@@ -1,8 +1,8 @@
 import { ILogGroup, LogGroup } from "aws-cdk-lib/aws-logs";
 import { Construct } from "constructs";
 import { IVpc } from "aws-cdk-lib/aws-ec2";
-import { AccessPoint, FileSystem, PerformanceMode } from "aws-cdk-lib/aws-efs";
-import { RemovalPolicy } from "aws-cdk-lib";
+import { AccessPoint, FileSystem, PerformanceMode, ThroughputMode } from "aws-cdk-lib/aws-efs";
+import { RemovalPolicy, Size } from "aws-cdk-lib";
 import { MountPoint, Volume } from "aws-cdk-lib/aws-ecs";
 
 export interface EngineProps {
@@ -59,9 +59,11 @@ export class Engine extends Construct {
     });
   }
 
-  protected createFileSystem(vpc: IVpc): FileSystem {
+  protected createFileSystem(vpc: IVpc, iops: Size): FileSystem {
     return new FileSystem(this, "FileSystem", {
       vpc: vpc,
+      provisionedThroughputPerSecond: iops,
+      throughputMode: ThroughputMode.PROVISIONED,
       encrypted: true,
       performanceMode: PerformanceMode.MAX_IO,
       removalPolicy: RemovalPolicy.DESTROY,
--- a/packages/cdk/lib/constructs/engines/engine.ts
+++ b/packages/cdk/lib/constructs/engines/engine.ts
@@ -59,11 +59,21 @@ export class Engine extends Construct {
     });
   }
 
-  protected createFileSystem(vpc: IVpc, iops: Size): FileSystem {
+  protected createFileSystemIOPS(vpc: IVpc, iops: Size): FileSystem {
     return new FileSystem(this, "FileSystem", {
       vpc: vpc,
       provisionedThroughputPerSecond: iops,
       throughputMode: ThroughputMode.PROVISIONED,
+      encrypted: true,
+      performanceMode: PerformanceMode.MAX_IO,
+      removalPolicy: RemovalPolicy.DESTROY,
+    });
+  }
+
+  protected createFileSystemDefaultThroughput(vpc: IVpc): FileSystem {
+    return new FileSystem(this, "FileSystem", {
+      vpc: vpc,
+      throughputMode: ThroughputMode.BURSTING,
       encrypted: true,
       performanceMode: PerformanceMode.MAX_IO,
       removalPolicy: RemovalPolicy.DESTROY,
--- a/packages/cdk/lib/env/context-app-parameters.ts
+++ b/packages/cdk/lib/env/context-app-parameters.ts
@@ -46,6 +46,10 @@ export class ContextAppParameters {
    */
   public readonly engineName: string;
   /**
+   * Name of the filesystem to use.
+   */
+  public readonly filesystemType?: string;
+  /**
    * Name of the engine ECR image.
    */
   public readonly engineDesignation: string;
@@ -98,6 +102,7 @@ export class ContextAppParameters {
     this.readWriteBucketArns = getEnvStringListOrDefault(node, "READ_WRITE_BUCKET_ARNS");
 
     this.engineName = getEnvString(node, "ENGINE_NAME");
+    this.filesystemType = getEnvStringOrDefault(node, "FILESYSTEM_TYPE", this.getDefaultFilesystem());
     this.engineDesignation = getEnvString(node, "ENGINE_DESIGNATION");
     this.engineHealthCheckPath = getEnvStringOrDefault(node, "ENGINE_HEALTH_CHECK_PATH", "/engine/v1/status")!;
     this.callCachingEnabled = getEnvBoolOrDefault(node, "CALL_CACHING_ENABLED", true)!;
@@ -150,5 +155,23 @@ export class ContextAppParameters {
         ...additionalEnvVars,
       },
     };
+  }
+
+  public getDefaultFilesystem(): string {
+    let defFilesystem: string;
+    switch (this.engineName) {
+      case "cromwell":
+        defFilesystem = "S3";
+        break;
+      case "nextflow":
+        defFilesystem = "S3";
+        break;
+      case "miniwdl":
+        defFilesystem = "EFS";
+        break;
+      default:
+        throw Error(`Engine '${this.engineName}' is not supported`);
+    }
+    return defFilesystem;
   }
 }
--- a/packages/cdk/lib/env/context-app-parameters.ts
+++ b/packages/cdk/lib/env/context-app-parameters.ts
@@ -46,9 +46,13 @@ export class ContextAppParameters {
    */
   public readonly engineName: string;
   /**
-   * Name of the filesystem to use.
+   * Name of the filesystem type to use (e.g. EFS, S3).
    */
   public readonly filesystemType?: string;
+  /**
+   * Name of the filesystem type to use (e.g. EFS, S3).
+   */
+  public readonly fsProvisionedThroughput?: number;
   /**
    * Name of the engine ECR image.
    */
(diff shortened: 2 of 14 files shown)


packages/cli/internal/pkg/cli/spec/project_test.go
```diff
@@ -86,13 +94,27 @@ schemaVersion: 0
 					"ctx1": {
 						MaxVCpus: 256,
 						Engines: []Engine{
-							{Type: "wdl", Engine: "miniwdl"},
+							{
+								Type:   "wdl",
+								Engine: "miniwdl",
+								Filesystem: Filesystem{
+									FSType:        "S3",
```
markjschreiber, line 101: Should this be EFS?

packages/cli/internal/pkg/cli/spec/project_test.go
```diff
@@ -122,11 +144,15 @@ contexts:
         engines:
             - type: wdl
               engine: miniwdl
+              filesystem:
+                fsType: S3
```
markjschreiber, line 148: EFS?

packages/cli/internal/pkg/cli/spec/project_test.go
```diff
@@ -168,6 +194,27 @@ contexts:
...
+	const yamlStr = `
+name: DefaultTest
+schemaVersion: 1
+contexts:
+    context:
+        engines:
+            - type: wdl
+              engine: cromwell
+              filesystem:
+                fsType: EFS
```
markjschreiber, line 207: Cromwell can't use EFS. This test should probably fail.

packages/cli/internal/pkg/cli/spec/project_test.go
```diff
@@ -187,12 +234,26 @@ func TestGetContext(t *testing.T) {
 					Contexts: map[string]Context{
 						"ctx1": {
 							Engines: []Engine{
-								{Type: "wdl", Engine: "miniwdl"},
+								{
+									Type:   "wdl",
+									Engine: "miniwdl",
+									Filesystem: Filesystem{
+										FSType:        "S3",
```
markjschreiber, line 241: miniwdl uses EFS

packages/cli/internal/pkg/cli/spec/project_test.go
```diff
@@ -187,12 +234,26 @@ func TestGetContext(t *testing.T) {
...
 						},
 						"ctx2": {
 							Engines: []Engine{
-								{Type: "nextflow", Engine: "nextflow"},
+								{
+									Type:   "nextflow",
+									Engine: "nextflow",
+									Filesystem: Filesystem{
+										FSType:        "S3",
+										Configuration: FSConfig{FSProvisionedThroughput: 0},
```
markjschreiber, line 254: S3 doesn't have a provisioned throughput. We should prevent this configuration from being an option with S3

packages/cli/internal/pkg/cli/spec/project_test.go
```diff
@@ -209,12 +270,26 @@ func TestGetContext(t *testing.T) {
 					Contexts: map[string]Context{
 						"ctx1": {
 							Engines: []Engine{
-								{Type: "wdl", Engine: "miniwdl"},
+								{
+									Type:   "wdl",
+									Engine: "miniwdl",
+									Filesystem: Filesystem{
+										FSType:        "S3",
```
markjschreiber, line 277: EFS

packages/cli/internal/pkg/cli/spec/project_test.go
```diff
@@ -223,7 +298,14 @@ func TestGetContext(t *testing.T) {
 			},
 			expectedContext: Context{
 				Engines: []Engine{
-					{Type: "wdl", Engine: "miniwdl"},
+					{
+						Type:   "wdl",
+						Engine: "miniwdl",
+						Filesystem: Filesystem{
+							FSType:        "S3",
```
markjschreiber, line 305: EFS

packages/cdk/lib/env/context-app-parameters.ts
```diff
@@ -45,6 +45,10 @@ export class ContextAppParameters {
    * Name of the engine to run.
    */
   public readonly engineName: string;
+  /**
+   * Name of the filesystem to use.
```
markjschreiber, line 49: Name of the filesystem type to use (e.g. EFS, S3)

packages/cdk/lib/env/context-app-parameters.ts
```diff
@@ -151,4 +156,22 @@ export class ContextAppParameters {
...
+      case "cromwell":
+        defFilesystem = "S3";
+        break;
+      case "nextflow":
+        defFilesystem = "S3";
+        break;
+      case "miniwdl":
+        defFilesystem = "EFS";
+        break;
+      default:
```
markjschreiber, line 172: Ask Elliot if he wants to add snakemake to this swith

markjschreiber:
Can you check and see if it's possible to enforce values for engine and filesystem in the `packages/cli/internal/pkg/cli/spec/project_schema.json`. For example, currently `cromwell` and `nextflow` must use s3 and `miniwdl` must use EFS.

You will also need to add logic so that the fileystem information is passed through to the cdk that generates the filesystem so that the resulting EFS volume has the desired provisioned IOPs.

packages/cli/internal/pkg/cli/spec/context.go
```diff
@@ -1,19 +1,46 @@
...
 type Context struct {
 	InstanceTypes        []string `yaml:"instanceTypes,omitempty"`
 	RequestSpotInstances bool     `yaml:"requestSpotInstances,omitempty"`
 	MaxVCpus             int      `yaml:"maxVCpus,omitempty"`
 	Engines              []Engine `yaml:"engines"`
 }
 
+func (filesystem *Filesystem) UnmarshalYAML(unmarshal func(interface{}) error) error {
+
+	type defValFilesystem Filesystem
```
tneely, line 29: Why are you using a type alias here? You should be able to just use `Filesystem{Configuration:...}`

packages/cli/internal/pkg/cli/spec/context.go
```diff
@@ -1,19 +1,46 @@
...
+	if err := unmarshal(&defaults); err != nil {
+		return err
+	}
+
+	*filesystem = Filesystem(defaults)
+	switch filesystem.FSType {
+	case "S3", "EFS", "":
+		return nil
+	default:
+		return fmt.Errorf("filesystem %s is invalid. Options are `S3` and `EFS`", filesystem.FSType)
```
tneely, line 40: nit: can you wrap the string variable in single quotes? It's generally easy to read/understand

```
fmt.Errorf("filesystem '%s' is invalid. Options are `S3` and `EFS`", filesystem.FSType)
```

packages/cli/internal/pkg/cli/spec/project_test.go
```diff
@@ -86,13 +94,27 @@ schemaVersion: 0
...
 					"ctx2": {
 						MaxVCpus: 256,
 						Engines: []Engine{
-							{Type: "nextflow", Engine: "nextflow"},
+							{
+								Type:   "nextflow",
+								Engine: "nextflow",
+								Filesystem: Filesystem{
+									FSType:        "S3",
+									Configuration: FSConfig{FSProvisionedThroughput: 0},
```
tneely, line 115: Can you add a test for when `FSProvisionedThroughput` isn't the default value?

packages/cli/internal/pkg/cli/spec/project_test.go
```diff
@@ -61,6 +67,8 @@ contexts:
         engines:
             - type: wdl
               engine: cromwell
+              filesystem:
+                fsType: S3
```
tneely, line 71: Do we have a test for when `filesystem` is not defined?

tneely (changes requested):

packages/cdk/lib/env/context-app-parameters.ts
elliot-smith: Yes, snakemake requires EFS so it would need to be here as well

packages/cdk/lib/constructs/engines/engine.ts
```diff
@@ -14,4 +17,56 @@ export class Engine extends Construct {
...
+    return new AccessPoint(this, "AccessPoint", {
+      fileSystem: fileSystem,
+      posixUser: {
+        uid: "0",
+        gid: "0",
+      },
+    });
+  }
+
+  protected createAccessPointcreateFileSystem(vpc: IVpc): FileSystem {
```
markjschreiber, line 53: Should this just be `createFileSystem` It doesn't create an access point unless you meant to call that method internally?

packages/cdk/lib/env/context-app-parameters.ts
```diff
@@ -45,6 +45,14 @@ export class ContextAppParameters {
    * Name of the engine to run.
    */
   public readonly engineName: string;
+  /**
+   * Name of the filesystem type to use (e.g. EFS, S3).
+   */
+  public readonly filesystemType?: string;
+  /**
+   * Name of the filesystem type to use (e.g. EFS, S3).
```
markjschreiber, line 53: Amount of provisioned IOPs?

packages/cdk/lib/types/engine-options.ts
```diff
@@ -15,6 +16,10 @@ export interface EngineOptions {
    * VPC to run resources in.
    */
   readonly vpc: IVpc;
+  /**
+   * Filesystem provisioned throughput to use for EFS.
+   */
+  readonly iops: Size;
```
markjschreiber, line 22: I think this should be an optional parameter. When the filesystem is S3 you can't supply iops and when the filesystem is EFS then iops is an option.

markjschreiber:
Looking good. General comments:

- Does the code handle the case where EFS is specified as the filesystem but provisioned IOPs is not (meaning you get the default MaxIO with burst mode). Provisioned IOPs is optional.
- There seems to be a lot of content from the snakemake implementation. Possibly a rebase on top of HEAD would generate a cleaner commit?

packages/cdk/lib/stacks/context-stack.ts
```diff
@@ -25,18 +26,33 @@ export class ContextStack extends Stack {
 
     const { contextParameters } = props;
     const { engineName } = contextParameters;
+    const { filesystemType } = contextParameters;
+    const { fsProvisionedThroughput } = contextParameters;
+    this.iops = Size.mebibytes(fsProvisionedThroughput!);
 
     switch (engineName) {
       case ENGINE_CROMWELL:
+        if (filesystemType != "S3") {
```
tneely, line 35: I'm a bit confused here. What's the point of exposing `filesystemType` if we error on any non-default values?

samwaln: I think the plan here is to have other filesystem types in the future.

tneely (approved):

PR #315, opened 2022-02-10
fix: Fix project data path prefix and docs

markjschreiber:
**Description of Changes**

corrected prefix data paths to include /* at the end so IAM permissions are correctly constructed.

* inspect IAM roles generated

**Checklist**

----

### DIFF
--- a/site/content/en/docs/Concepts/projects.md
+++ b/site/content/en/docs/Concepts/projects.md
@@ -94,7 +94,7 @@ data:
 ```
 
 You can use S3 prefixes to be more restrictive about access to data. For example, if you want to allow access to the 
-`foo` folder or `my-bucket` and it's sub-folders you would declare the location as:
+`foo` folder of `my-bucket` and it's sub-folders you would declare the location as:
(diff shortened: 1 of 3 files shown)


site/content/en/docs/Concepts/projects.md
```diff
@@ -92,6 +93,21 @@ data:
     readOnly: true
 ```
 
+You can use S3 prefixes to be more restrictive about access to data. For example, if you want to allow access to the 
+`foo` folder or `my-bucket` and it's sub-folders you would declare the location as:
```
tneely, line 97: Should this be "of" not "or"? 

"...access to the `foo` folder **_of_** `my-bucket`..."?

elliot-smith (approved):

site/content/en/docs/Concepts/projects.md
markjschreiber: Good find, fixed in latest commit

tneely (approved):

Issue #314, opened 2022-02-09
AGC thinks a WDL `Log` object's `exit_code` should be a string, but the spec defines it as an integer


adamnovak:
[//]: #  (Please do not include your AWS account or credentials as part of the bug request.)

**Describe the Bug**

[//]: #  (Please describe the behavior that you are seeing with as many details as possible including logs from AGC, CloudWatch snippets and CloudFormation errors if relevant.)

I reported this as https://github.com/DataBiosphere/toil/issues/4030, but according to [the spec](https://github.com/ga4gh/workflow-execution-service-schemas/blob/c3b19854240c4fcbaf3483e22b19db0a918a7ee5/openapi/workflow_execution_service.swagger.yaml#L409-L412) I think Toil is right to send an integer here and AGC is wrong to require a string.

**Steps to Reproduce**

[//]: #  (Include a list of steps that can be used to reproduce the bug.)

1. Use AGC with Toil engine: https://github.com/adamnovak/amazon-genomics-cli/commit/d831ee9c346a545ea63b11a82da26b3cc8ccecad
2. Set up a Toil engine Docker container for Toil https://github.com/DataBiosphere/toil/commit/fc8e0e2d63ed5d2db609654ddd6a0c3e86d8144f, using the Toil Dockerfile in AGC, and put it in ECR.
3. Set up a Toil engine context using the CWL demo context that uses Toil
4. Run a workflow
5. Retrieve the workflow logs with `agc workflow output`

**Relevant Logs**

[//]: #  (The logs that where obtained by running the commands here. Please run the command with "-v" so that we can see the verbose logs.)

```
2022-02-01T15:17:07-08:00 ✘   error="json: cannot unmarshal number into Go struct field Log.run_log.exit_code of type string"
Error: an error occurred invoking 'workflow output'
with variables: {runId:b4b54c271ab1453f88c7912cf8c757b9}
caused by: json: cannot unmarshal number into Go struct field Log.run_log.exit_code of type string
```

**Expected Behavior**

[//]: #  (A clear and concise description of what you expected to happen.)

AGC should be able to unmarshall a run log exit code that comes in as an unquoted integer, as the spec defines it as an integer.

**Actual Behavior**

[//]: #  (A clear and concise description of what actually happened.)

AGC fails at deserialization unless the exit code is quoted as a string in the WES JSON response.

**Screenshots**

[//]: #  (If applicable, add screenshots to help explain your problem.)

N/A

**Additional Context**

[...]

PR #317, opened 2022-02-10
feat: Toil engine for CWL

adamnovak:
Issue #, if available: N/A

**Description of Changes**

This PR contains almost the minimum changes necessary to add Toil as an engine to run CWL.

It also adds the ability to override the Docker image names used when deploying a context (because this is necessary to actually use the Toil engine with Docker images with names in one's own namespace), and the ability to pass only relevant Docker images to the CDK (because otherwise the CDK commands grow with each engine and each new engine has to touch the part of the tests that counts how long the commands are).

I can't actually run this as submitted, because it doesn't have #231.

You can't actually read the logs of the successful workflow out, because pulling logs over WES is a separate feature (partly blocked by #314), and I'm going to include what I have for that in a separate PR.

But I tested commit 822e51af7a7316282d73dc9f99d878dbd27fe85b and [I was able to run the included hello world workflow](https://github.com/DataBiosphere/toil/issues/3790#issuecomment-1035480921).

**Checklist**

----


adamnovak:
Looks like my attempt to resolve the merge conflict on Github broke the tests; I will have to pull down the branch and fix the merge.

Any advice for how to deal with the CDK call argument count tests? My changes conflict with any changes in `main`, and the argument counts in `main` change quite often.

examples/demo-cwl-project/agc-project.yaml
```diff
@@ -0,0 +1,20 @@
+---
+name: Demo
```
markjschreiber, line 2: Can you use DemoCwl or similar so that the deployed artifacts have naming that is easier to follow in testing?

packages/cdk/lib/roles/policies/toil-batch-policy.ts
```diff
@@ -0,0 +1,33 @@
+import { PolicyDocument, PolicyStatement, Effect } from "aws-cdk-lib/aws-iam";
+
+export interface ToilBatchPolicyProps {
+  jobQueueArn: string;
+  // This is actually a pattern that matches all ARNs for potentially relevant
+  // definitions, since Toil makes its own definitions.
+  toilJobArn: string;
```
markjschreiber, line 7: toilJobDefinitionArnPattern would be a more meaningful name

packages/cdk/lib/roles/policies/toil-batch-policy.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export class ToilBatchPolicy extends PolicyDocument {
+  constructor(props: ToilBatchPolicyProps) {
+    super({
+      assignSids: true,
+      statements: [
+        new PolicyStatement({
+          effect: Effect.ALLOW,
+          actions: ["batch:DescribeJobDefinitions", "batch:ListJobs", "batch:DescribeJobs", "batch:DescribeJobQueues", "batch:DescribeComputeEnvironments"],
+          resources: ["*"],
```
markjschreiber, line 18: I think some of these can have a more limited scope. For example, you know the job queue so possibly `batch:DescribeJobQueues` can be limited to that arn. You may also be able to limit `batch:DescribeJobDefinitions` to the `toilJobDeinitionArnPattern`. You may have to split the policies to do this but in general the number of actions with `resources: ["*"]` should be as small as possible

packages/cdk/lib/roles/toil-job-role.ts
```diff
@@ -0,0 +1,50 @@
...
+      assumedBy: new ServicePrincipal("ecs-tasks.amazonaws.com"),
+      inlinePolicies: {
+        // TODO: Remove this when Toil no longer uses its own SimpleDB domains
+        ToilSimpleDBFullAccess: new PolicyDocument({
+          assignSids: true,
+          statements: [
+            new PolicyStatement({
+              effect: Effect.ALLOW,
+              actions: ["sdb:*"],
+              resources: ["*"],
```
markjschreiber, line 26: Is it possible to know this arn and restrict to it?

packages/cdk/lib/roles/toil-job-role.ts
```diff
@@ -0,0 +1,50 @@
...
+    super(scope, id, {
+      assumedBy: new ServicePrincipal("ecs-tasks.amazonaws.com"),
+      inlinePolicies: {
+        // TODO: Remove this when Toil no longer uses its own SimpleDB domains
+        ToilSimpleDBFullAccess: new PolicyDocument({
+          assignSids: true,
+          statements: [
+            new PolicyStatement({
+              effect: Effect.ALLOW,
+              actions: ["sdb:*"],
```
markjschreiber, line 25: Probably more comments on this later but why not use DynamoDB?

packages/cdk/lib/roles/toil-job-role.ts
```diff
@@ -0,0 +1,50 @@
...
+          ],
+        }),
+        // TODO: Remove this when Toil is taught to use AGC buckets to store
+        // its workflow state and doesn't need to make and destroy its own.
+        ToilS3FullAccess: new PolicyDocument({
+          assignSids: true,
+          statements: [
+            new PolicyStatement({
+              effect: Effect.ALLOW,
+              actions: ["s3:*"],
```
markjschreiber, line 37: This is far too permissive. It could be used to do literally anything to a customers buckets. This must be limited to a pattern or single bucket.

packages/cdk/lib/stacks/context-stack.ts
```diff
@@ -143,6 +178,19 @@ export class ContextStack extends Stack {
     };
   }
 
+  private getToilBatchProps(props: ContextStackProps) {
+    const commonBatchProps = this.getCommonBatchProps(props);
+    const { requestSpotInstances } = props.contextParameters;
+
+    return {
+      ...commonBatchProps,
+      // We only use one Batch from the stack for the Toil jobs. The server
```
markjschreiber, line 187: We only use one Batch compute environment and queue

packages/cdk/lib/stacks/engines/toil-engine-construct.ts
```diff
@@ -0,0 +1,104 @@
...
+import { LogGroup, ILogGroup } from "aws-cdk-lib/aws-logs";
+import { EngineOutputs, EngineConstruct } from "./engine-construct";
+import { ToilJobRole } from "../../roles/toil-job-role";
+import { ToilEngineRole } from "../../roles/toil-engine-role";
+import { IJobQueue } from "@aws-cdk/aws-batch-alpha";
+import { Construct } from "constructs";
+
+export interface ToilEngineConstructProps extends EngineOptions {
+  /**
+   * AWS Batch JobQueue to use for running workflows.
```
markjschreiber, line 18: Is it used for running workflows or workflow tasks (or both)?

packages/cdk/lib/stacks/engines/toil-engine-construct.ts
```diff
@@ -0,0 +1,104 @@
...
+   */
+  readonly jobQueue: IJobQueue;
+}
+
+export class ToilEngineConstruct extends EngineConstruct {
+  public readonly engine: SecureService;
+  public readonly adapterRole: IRole;
```
markjschreiber, line 25: If Toil can speak WES without needing to use a Lambda why is there an `adapterRole` and `adapterLogGroup`? Perhaps these need renaming?

packages/cdk/lib/stacks/engines/toil-engine-construct.ts
```diff
@@ -0,0 +1,104 @@
...
+      readWriteBucketArns: (params.readWriteBucketArns ?? []).concat(outputBucket.bucketArn),
+      policies: props.policyOptions,
+    });
+
+    // Make the container and pass it the ARN of the role to use for individual jobs.
+    const engineContainer = params.getEngineContainer(props.jobQueue.jobQueueArn, {
+      TOIL_AWS_BATCH_JOB_ROLE_ARN: this.jobRole.roleArn,
+    });
+
+    // TODO: Move log group creation into service construct and make it a property
```
markjschreiber, line 60: Is there anything blocking this from happening right now? If not either move it or remove the TODO

packages/cdk/lib/stacks/engines/toil-engine-construct.ts
```diff
@@ -0,0 +1,104 @@
...
+    this.engine = this.getEngineServiceDefinition(props.vpc, engineContainer, this.engineLogGroup);
+    // This is unused because we have no adapter, but a log group is required.
+    this.adapterLogGroup = new LogGroup(this, "AdapterLogGroup");
```
markjschreiber, line 63: Another way to think about the adapterLogGroup would be the LogGroup for WES logs. The name reflects the notion that all of our engines don't understand WES natively. Is there a way that WES related logging from Toil can go here?

packages/cdk/lib/stacks/engines/toil-engine-construct.ts
```diff
@@ -0,0 +1,104 @@
...
+      apiName: `${params.projectName}${params.contextName}${engineContainer.serviceName}ApiProxy`,
+      loadBalancer: this.engine.loadBalancer,
+      allowedAccountIds: [Aws.ACCOUNT_ID],
+    });
+  }
+
+  protected getOutputs(): EngineOutputs {
+    return {
+      accessLogGroup: this.apiProxy.accessLogGroup,
+      adapterLogGroup: this.adapterLogGroup,
```
markjschreiber, line 77: If you cannot make an `adapterLogGroup` from Toil then you could make this field of `EngineOutputs` optional?

packages/cli/internal/pkg/cli/context/manager.go
```diff
@@ -87,8 +87,16 @@ var showExecution = cdk.ShowExecution
 var silentExecution = cdk.SilentExecution
 
 func (m *Manager) getEnvironmentVars() []string {
+	// Different engines will need different environment variables to define
+	// their Docker images.
+	engine := m.contextEnv.EngineName
+	// Each engine has its own section in imageRefs, and for now we assume they
+	// all care about a WES adapter image.
```
markjschreiber, line 94: I don't think we should assume this. Toils is an example of an engine with no WES Adapter image so rather than faking it we should handle the edge case where there is no WES adapter

markjschreiber (changes requested):
Looking good. Some comments on improvements that could be made. Most important is paring back the `s3:*` on `*` permission. This won't fly with security or with account admins who need to install AGC.

We will also need some basic additions to the github pages docs. This really just means adding a `toil.md` file in `site/content/en/docs/Workflow engines`. You can follow the basic pattern of the other engine pages.

packages/cdk/lib/roles/toil-job-role.ts
adamnovak: Right now Toil has code that requires SimpleDB. Now that S3 is strongly consistent, we have [a project to remove the SimpleDB dependency and just use S3](https://github.com/DataBiosphere/toil/pull/3569), but it is not yet complete, so currently Toil needs SimpleDB to work.

packages/cdk/lib/roles/toil-job-role.ts
adamnovak: Toil is going to make a buckets and a SimpleDB domains with names starting with `toil-cwl-` for each workflow, for scratch. It also I believe uses a `toil` SimpleDB domain.

We probably should be able to grant permissions for just those prefixes.

This will cause trouble if the user wants to specify their own `--jobStore` and tell Toil to use a different bucket, but we don't necessarily have to support that.

packages/cdk/lib/stacks/engines/toil-engine-construct.ts
adamnovak: I left `adapterRole` and `adapterLogGroup` in under the impression that the type system required them to be there. Toil doesn't use them, and on looking at this again it seems like  `adapterRole` (now?) never makes it out to the base engine interface.

But [`EngineOutputs` always has an `adapterLogGroup` right now](https://github.com/aws/amazon-genomics-cli/blob/75603a860ca401b7be2d5d04c51248a227283d00/packages/cdk/lib/stacks/engines/engine-construct.ts#L12-L17), and the base `EngineConstruct` defines a method that has to return an `EngineOutputs`. I think to get rid of it I would need to not make `ToilEngineConstruct` an `EngineConstruct`, or else remove the field from the base `EngineOutputs` and duplicate it and any code that uses it in all the other engines. Or else add a new intermediate `AdaptedEngineOutputs` and `AdaptedEngineConstruct` and change the other engines to use those.

packages/cdk/lib/stacks/engines/toil-engine-construct.ts
adamnovak: This comment came along with the other code I [duplicated from the Cromwell engine](https://github.com/aws/amazon-genomics-cli/blob/75603a860ca401b7be2d5d04c51248a227283d00/packages/cdk/lib/stacks/engines/cromwell-engine-construct.ts#L52-L54).

I could probably jut drop it here. There won't be an adapter log group to move if I manage to defeat the type system and get rid of it.

packages/cdk/lib/stacks/engines/toil-engine-construct.ts
adamnovak: That would involve finding some kind of Python `logging` module backend that could send logs to a CloudWatch log group instead of standard error, and adding it to Toil, and configuring it to pick up messages from WES-related modules, I think. It seems like making the notion of an adapter optional would be a lot easier.

packages/cdk/lib/stacks/engines/toil-engine-construct.ts
adamnovak: Oh yeah, I suppose fields can be optional, maybe that is the right approach to take.

examples/demo-cwl-project/agc-project.yaml
adamnovak: I will change this to be unique.

packages/cdk/lib/roles/policies/toil-batch-policy.ts
adamnovak: This is just a copy-paste of the policy that Cromwell is using:

https://github.com/aws/amazon-genomics-cli/blob/8bda67492b3cf143ffd5a0a8931bbcdf3e04992d/packages/cdk/lib/roles/policies/cromwell-batch-policy.ts#L13-L17

Testing any change here is going to take something like an hour of turnaround time to tear down and redeploy the stack, so I'm very reluctant to undertake myself the project of determining what the minimum permission set to use Batch actually is.

I think instead of copy-pasting the code, I will try and base this off the Cromwell policy more directly, so that when the Cromwell permission set gets tightened, Toil will also be fixed.

packages/cdk/lib/roles/policies/toil-batch-policy.ts
adamnovak: However, a similar change may want to be made to `cromwellJobArn`, which is also a pattern for engine-managed job definitions: https://github.com/aws/amazon-genomics-cli/blob/8bda67492b3cf143ffd5a0a8931bbcdf3e04992d/packages/cdk/lib/roles/cromwell-engine-role.ts#L17-L26. I don't think that would make sense to do in this PR.

packages/cdk/lib/roles/toil-job-role.ts
adamnovak: The ARNs that Toil needs follow a pattern that I will restrict to.

packages/cdk/lib/stacks/engines/toil-engine-construct.ts
adamnovak: This is just for the tasks; I can note that.

packages/cli/internal/pkg/cli/context/manager.go
adamnovak: I can make it support passing the WES adapter for the engines that need it only.

packages/cli/internal/pkg/cli/context/manager_deploy_test.go
```diff
@@ -46,7 +46,7 @@ func TestManager_Deploy(t *testing.T) {
 				mockClients.ssmMock.EXPECT().GetCustomTags().Return(testTags)
 				mockClients.ecrClientMock.EXPECT().VerifyImageExists(environment.CommonImages["NEXTFLOW"]).Return(nil)
 				clearContext := mockClients.cdkMock.EXPECT().ClearContext(filepath.Join(testHomeDir, ".agc/cdk/apps/context")).Return(nil)
-				mockClients.cdkMock.EXPECT().DeployApp(filepath.Join(testHomeDir, ".agc/cdk/apps/context"), gomock.Len(42), testContextName3).After(clearContext).Return(mockClients.progressStream1, nil)
+				mockClients.cdkMock.EXPECT().DeployApp(filepath.Join(testHomeDir, ".agc/cdk/apps/context"), gomock.Len(30), testContextName3).After(clearContext).Return(mockClients.progressStream1, nil)
```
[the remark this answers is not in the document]
adamnovak: I'll do that.

packages/cli/internal/pkg/cli/project_init.go
```diff
@@ -125,7 +126,7 @@ func (o *initProjectOpts) validateProject() error {
 func BuildProjectInitCommand() *cobra.Command {
 	vars := initProjectVars{}
 	cmd := &cobra.Command{
-		Use:   "init project_name --workflow-type {wdl|nextflow|snakemake}",
+		Use:   "init project_name --workflow-type {cwl|wdl|nextflow|snakemake}",
```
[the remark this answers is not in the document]
adamnovak: I'll do that.

packages/engines/toil/Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+    && rm -rf /tmp/aws*
+
+##### MODIFY #######
+## In this area install your new engine into the container as well as any requirements for that engine.
+## Dockerfile documentation is found here: https://docs.docker.com/engine/reference/builder/
+
+# Add rabbitmq repository
+ADD rabbitmq.repo /etc/yum.repos.d/rabbitmq.repo
+
+# Sadly pre-importing keys doesn't seem to save any time whan we use yum later, so don't so it.
```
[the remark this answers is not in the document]
adamnovak: I will fix this.

packages/engines/toil/README.md
```diff
@@ -0,0 +1,48 @@
+## Toil AWS Mirror
+
+A Toil mono-container WES server for use with Amazon AGC.
```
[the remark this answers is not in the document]
adamnovak: I'll call it that.

packages/engines/toil/README.md
```diff
@@ -0,0 +1,48 @@
...
+```bash
+docker exec -ti "$(docker ps | grep adamnovak/toil-agc | rev | cut -f1 -d' ' | rev)" /bin/bash
+```
+
+### Deploying
+
+To push this to an Amazon ECR repo, where AGC can get at it, you can do something like:
+
+```bash
+aws ecr get-login-password --region us-west-2 | docker login --username AWS --password-stdin 318423852362.dkr.ecr.us-west-2.amazonaws.com
```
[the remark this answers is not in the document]
adamnovak: I can do this, although then I won't be able to copy-paste this and execute it anymore.

packages/engines/toil/README.md
```diff
@@ -0,0 +1,48 @@
...
+docker build -t adamnovak/toil-agc .
+docker tag adamnovak/toil-agc:latest 318423852362.dkr.ecr.us-west-2.amazonaws.com/adamnovak/toil-agc:latest
```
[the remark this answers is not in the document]
adamnovak: I will do this.

adamnovak:
@markjschreiber I've added some Toil documentation like you asked for. I generated the documentation and it looks right to me.

adamnovak:
As discussed at the meeting today, the AGC project is unwilling to take this without locking down the IAM actions granted to the Toil engine, beyond just restricting access to certain buckets/domains.

I've added what I think is going to be close to the least-privilege set of permissions. I got these by looking at the full lists of possible permissions, and keeping the ones that appeared to correspond to Boto3 Client calls we make, or to Boto3 Resource objects or collections we access or things we do to those objects and collections.

I still need to:
1. Test to make sure that these permissions are sufficient, at least for an example workflow.
2. Somehow test to see if they are minimal, or at least close enough to minimal to be getting on with.

I don't think we'll be able to guarantee actual minimality with a reasonable time investment. We'd need to put together a workflow set that exercises every Boto3 call in Toil's job store, and try it once per permission, with that permission left out. Or we'd need to inventory every call into Boto3, trace them through the Boto3 Resource code to the API calls that will be made (accounting for any laziness), and then determine what IAM actions those API calls need allowed, either from the documentation or by experiment.

For example, Toil at one point says:

```
bucket.objects.all().delete()
```

Presumably we need `s3:DeleteObject` for that, and something for listing the bucket objects. We'd need to figure out if the permission needed for listing here is `s3:ListBucket` or `s3:ListObjects` if we really wanted a *minimal* set of permissions, and the answer (or indeed the distinction) doesn't appear to be documented.

It was mentioned that CloudWatch can log API operations, but I'm not sure that's the same as logging IAM action checks hit.

adamnovak:
OK, I had to make some changes to the permission set I initially proposed, but I've now tested this and it can successfully run the example `nontrivial` CWL workflow.

markjschreiber:
A rebased version of this PR has been merged to HEAD.

PR #352, opened 2022-03-01
feat: Add optionsFile flag

samwaln:
Issue #, if available: N/A

**Description of Changes**
Add flags for optionFileUrl and engineOptions. The first of these is for use with engines that run in server mode. The latter is for use with head node engines. Warnings are logged if the flag is not valid for the engine type and the override will be ignored.

Ran workflows with various options. Example: 'agc workflow run words-with-vowels -c myContext -a workflows/words/words-with-vowels.inputs.json -o workflows/words/words-with-vowels.optionFile.json'.
Unit tests to check that values are set.

**Checklist**

----


packages/cli/internal/pkg/storage/option_file_client.go
```diff
@@ -0,0 +1,123 @@
...
+
+	return nil
+}
+
+func (oc *OptionInstance) UpdateOptionFile(initialProjectDirectory string, optionFile interface{}, bucketName string, baseS3Key string, fileLocation string) (interface{}, error) {
+	var updatedOptionReferenceFile interface{}
+	var optionReference string
+	switch typedValue := optionFile.(type) {
+	case string:
+		updatedReference, err := oc.uploadReferenceToS3(typedValue, initialProjectDirectory, bucketName, baseS3Key)
```
ShubhamChaturvedi7, line 57: Do we have files/urls as keys in the optionsFile? I though it was all values/params for the engine.

packages/cli/internal/pkg/storage/option_file_client.go
```diff
@@ -0,0 +1,123 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		updatedOptionReferenceFile = updatedReference
+	case []interface{}:
```
ShubhamChaturvedi7, line 63: Extending the comment above, why do we need this?

packages/cli/internal/pkg/storage/option_file_client_test.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+func (oc *OptionClientTestSuite) TestUpdateOptionsFile_Success() {
+	optionFileUrl := testFileUrl
+	expectedUpdatedOptionFile := testFileUrl
+	oc.mockOs.EXPECT().Stat(initialProjectDirectory+"/"+testFile1).AnyTimes().Return(nil, nil)
+	expectedErr := errors.New("FileNotFound")
+	oc.mockOs.EXPECT().Stat(gomock.Any()).AnyTimes().Return(nil, expectedErr)
+	oc.mockJson.EXPECT().Marshal(gomock.Any()).Return(nil, nil)
+	oc.mockFileWriter.EXPECT().WriteFile(gomock.Any(), gomock.Any(), gomock.Any()).Return(nil)
```
ShubhamChaturvedi7, line 54: Let's avoid gomock.Any() wherever possible.

packages/cli/internal/pkg/cli/workflow/manager.go
```diff
@@ -365,6 +372,56 @@ func (m *Manager) uploadInputFileToS3(inputKey inputKey, fileUrl inputUrl) input
...
+		return
+	}
+	m.optionFile = optionFile
+}
+
+func (m *Manager) uploadOptionFileToS3() {
+	if m.err != nil || m.optionFile == nil {
+		return
+	}
+	objectKey := awsresources.RenderBucketDataKey(m.projectSpec.Name, m.userId)
```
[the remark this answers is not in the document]
samwaln: RenderBucketDataKey takes a projectName and userId

packages/cli/internal/pkg/cli/workflow/workflow_run.go
```diff
@@ -22,11 +22,15 @@ func (m *Manager) RunWorkflow(contextName, workflowName, argumentsUrl string) (s
 	m.readInput(argumentsUrl)
 	m.uploadInputsToS3()
 	m.parseInputToArguments()
+	m.readOptionFile(optionFileUrl)
+	m.uploadOptionFileToS3()
+	m.readEngineOptions(engineOptions)
```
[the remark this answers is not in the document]
samwaln: I'm going to make a separate issue for this because it looks like this is not actually an equivalent to optionFileUrl for Cromwell

site/content/en/docs/Concepts/workflows.md
```diff
@@ -140,6 +140,27 @@ agc workflow run my-workflow --args inputs.json
...
+```shell
+agc workflow run my-workflow --optionFileUrl optionFile.json
+```
+
+If the option file references a local file, it will be synced with S3 and the file in S3 will be used when the workflow
+instance is run. This is only for use with engines that run in server mode (e.g. Cromwell). It will be ignored if not using an engine  running in server mode.
+
+#### `workflow engineOptions`
+
+An additional property 'engineOptions' can be provided using the 'e' or '--engineOptions' flag. For example:
```
tneely, line 156: Since `engineOptions` is part of `workflowEngineParams`, specifying both flags is currently undefined behavior. Presumably we want `--engineOptions` to override the value used by same named key of the file specified with  `--optionFileUrl`, but we should call out that behavior so customers know how they interact together

site/content/en/docs/Concepts/workflows.md
```diff
@@ -140,6 +140,27 @@ agc workflow run my-workflow --args inputs.json
 If the inputs file references local files, these will be synced with S3 and those files in S3 will be used when the workflow
 instance is run.
 
+#### `workflow optionFileUrl`
+
+An additional optionFileUrl can be provided using the 'o' or '--optionFileUrl' flag. For example:
+
+```shell
+agc workflow run my-workflow --optionFileUrl optionFile.json
```
tneely, line 148: Have we defined anywhere what `optionFile.json` looks like?

tneely (changes requested):

packages/cli/internal/pkg/cli/workflow/manager.go
```diff
@@ -65,19 +65,23 @@ type s3Props struct {
...
+	workflowEngine       string
+	parsedSourceURL      *url.URL
+	isLocal              bool
+	path                 string
+	packPath             string
+	workflowUrl          string
+	inputUrl             string
+	input                Input
+	optionFileUrl        string
+	optionFile           map[string]string
```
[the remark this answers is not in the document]
ShubhamChaturvedi7: These are for the options for the server mode engines while engineOptions are for the head node engines

packages/cli/internal/pkg/cli/workflow/manager.go
```diff
@@ -458,6 +495,31 @@ func (m *Manager) setWorkflowParameters() {
...
+		return
+	}
+	m.workflowEngineParams = make(map[string]string)
+	if m.optionFileUrl == "" {
+		return
+	}
+	if m.optionFile != nil {
+		if m.workflowEngine == "nextflow" || m.workflowEngine == "miniwdl" {
+			log.Debug().Msgf("optionFile flag can only be used with head node engines")
+			return
```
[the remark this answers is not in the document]
ShubhamChaturvedi7: Yeah but the idea for the feature is to ignore the options But log this at the DEBUG level

ShubhamChaturvedi7 (changes requested):

packages/cli/internal/pkg/cli/workflow/manager.go
```diff
@@ -458,6 +495,31 @@ func (m *Manager) setWorkflowParameters() {
...
+	if m.err != nil {
+		return
+	}
+	m.workflowEngineParams = make(map[string]string)
+	if m.optionFileUrl == "" {
+		return
+	}
+	if m.optionFile != nil {
+		if m.workflowEngine == "nextflow" || m.workflowEngine == "miniwdl" {
+			log.Debug().Msgf("optionFile flag can only be used with head node engines")
```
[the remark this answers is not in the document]
samwaln: Good catch, it should be "cannot be used with"

site/content/en/docs/Concepts/workflows.md
samwaln: This is what the understanding is at the moment, but I'm creating a new issue to look into this more since it looks like they are not actually equivalents for server mode and head node engines.

packages/cli/internal/pkg/cli/workflow/manager.go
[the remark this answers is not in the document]
tneely: Why? We try to be transparent to our users and avoid magic in favor of being explicit.

packages/cli/internal/pkg/cli/workflow/manager.go
```diff
@@ -436,6 +475,31 @@ func (m *Manager) setWorkflowParameters() {
...
+	if m.err != nil {
+		return
+	}
+	m.workflowEngineParams = make(map[string]string)
+	if m.optionFileUrl == "" {
+		return
+	}
+	if m.options != nil {
+		if m.workflowEngine == "nextflow" || m.workflowEngine == "miniwdl" {
+			log.Error().Msgf("optionFile flag cannot be used with head node engines")
```
tneely, line 488: Please return an error (`m.err = ...`) rather than log one and continue

packages/cli/internal/pkg/cli/workflow/manager.go
```diff
@@ -436,6 +475,31 @@ func (m *Manager) setWorkflowParameters() {
...
+		m.workflowEngineParams = m.options
+	}
+
+	namePattern := fmt.Sprintf("%s_*", filepath.Base(m.optionFileUrl))
+	optionFileName, err := writeToTmp(namePattern, m.optionFileUrl)
+	if err != nil {
+		m.err = err
+		return
+	}
+	m.attachments = append(m.attachments, optionFileName)
```
tneely, line 500: How does this work? We're both passing the option file as an attachment and as a map of options. It feels redundant - shouldn't only one would be needed?

packages/cli/internal/pkg/cli/workflow/workflow_run_test.go
```diff
@@ -261,7 +264,35 @@ func (s *WorkflowRunTestSuite) TestRunWorkflow_LocalFile_NoArgs() {
...
 	s.mockOs.EXPECT().Remove(testCompressedTmpPath).After(uploadCall).Return(nil)
 
-	actualId, err := s.manager.RunWorkflow(testContext1Name, testLocalWorkflowName, "")
+	actualId, err := s.manager.RunWorkflow(testContext1Name, testLocalWorkflowName, "", "", "")
+	if s.Assert().NoError(err) {
+		s.Assert().Equal(testRun1Id, actualId)
+	}
+}
+
+func (s *WorkflowRunTestSuite) TestRunWorkflow_LocalFile_OptionsFile() {
```
tneely, line 273: Can you please add a test for `engineOptions` as well?

tneely (changes requested):

packages/cli/internal/pkg/cli/workflow/manager.go
samwaln: We don't need to save it as an attachment

tneely (dismissed):

packages/cli/internal/pkg/cli/workflow/manager.go
```diff
@@ -436,6 +475,23 @@ func (m *Manager) setWorkflowParameters() {
...
+	m.workflowEngineParams = make(map[string]string)
+	if m.optionFileUrl == "" {
+		return
+	}
+	if m.options != nil {
+		if m.workflowEngine == "nextflow" || m.workflowEngine == "miniwdl" {
+			m.err = fmt.Errorf("optionFile flag cannot be used with head node engines")
+			return
+		}
+		m.workflowEngineParams = m.options
```
ShubhamChaturvedi7, line 491: How do we pass this to the actual engine?

packages/cli/internal/pkg/cli/workflow/workflow_run.go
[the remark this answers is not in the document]
ShubhamChaturvedi7: In that case, let's remove all engineOptions related code.

packages/cli/internal/pkg/cli/workflow/workflow_run_test.go
```diff
@@ -160,17 +167,33 @@ func (s *WorkflowRunTestSuite) BeforeTest(_, _ string) {
...
-	s.wfInstance = ddb.WorkflowInstance{
+	s.wfInstance1 = ddb.WorkflowInstance{
 		RunId:        testRun1Id,
 		WorkflowName: testLocalWorkflowName,
 		ContextName:  testContext1Name,
 		ProjectName:  testProjectName,
 		UserId:       testUserId,
 	}
 
+	s.wfInstance2 = ddb.WorkflowInstance{
```
ShubhamChaturvedi7, line 189: Why do we need another WorkflowInstance?

ShubhamChaturvedi7 (commented):
1. Let's remove any engineOptions related code if we aren't implementing it in this PR.
2. There are few pending comments, please do respond even if we aren't planning to address them right away.
3. Did we test that optionsFile is actually being passed to the engine?

packages/cli/internal/pkg/cli/workflow/workflow_run_test.go
samwaln: We need a different context to test engineOptions since it can only be used with head node engines.

packages/cli/internal/pkg/cli/workflow/manager.go
samwaln: The parameter is set in `runWorkflow()` with `option.WorkflowEngineParams(m.workflowEngineParams))`

packages/cli/internal/pkg/cli/workflow/manager.go
```diff
@@ -365,6 +372,56 @@ func (m *Manager) uploadInputFileToS3(inputKey inputKey, fileUrl inputUrl) input
...
+	}
+	m.optionFile = updatedOption
+}
+
+func (m *Manager) readEngineOptions(engineOptions string) {
+	if m.err != nil || engineOptions == "" {
+		return
+	}
+	log.Debug().Msgf("engineOptions override: %s", engineOptions)
+	if m.workflowEngine == "cromwell" {
```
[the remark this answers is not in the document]
samwaln: An enum here would not be that clean since there are only three engines and two types.

packages/cli/internal/pkg/cli/workflow/workflow_run.go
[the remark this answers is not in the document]
samwaln: This is implemented as we understand engineOptions to work. The new issue will be to investigate this because from my research, this is not quite right.

ShubhamChaturvedi7 (approved):

tneely (approved):

PR #353, opened 2022-03-01
feat: Record command run and show in agc workflow status

samwaln:
Issue #, if available: N/A

**Description of Changes**

This change updates the output of 'agc workflow status' to show the command run. This will show any overrides that were used and any options.

Updated unit tests to include "Request" field. This checks the status of an instance.
Ran manual testing with 'agc workflow status' to check output.

**Checklist**

----

### DIFF
--- a/packages/cli/internal/pkg/cli/workflow/manager.go
+++ b/packages/cli/internal/pkg/cli/workflow/manager.go
@@ -667,7 +667,11 @@ func (m *Manager) updateInProject(instance *InstanceSummary) {
 }
 
 func (m *Manager) setRequest(instance *InstanceSummary) {
-	if m.err != nil || instance == nil || instance.Request != "" {
+	if m.err != nil || instance == nil {
+		return
+	}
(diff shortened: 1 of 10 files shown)


packages/cli/internal/pkg/cli/workflow/manager.go
```diff
@@ -663,6 +666,22 @@ func (m *Manager) updateInProject(instance *InstanceSummary) {
 	_, instance.InProject = m.projectSpec.Workflows[instance.WorkflowName]
 }
 
+func (m *Manager) setRequest(instance *InstanceSummary) {
+	if m.err != nil || instance == nil || instance.Request != "" {
```
tneely, line 670: If `instance` is already null or `Request` already set, then there's probably something wrong in our code. Rather than silently fail to set the request, we should at the very least log a warning that something is wrong.

tneely (changes requested):

tneely (approved):

PR #365, opened 2022-03-08
feat: ensures vpcId is recalled (if available) on account_activate

OhCloud:
Issue #, if available:

**Description of Changes**
Previously, `account activate` did not recall any existing vpcIds and users had to add it in manually. Now, if a vpcId exists, it will recall and use it in-place of generating a new one. 

Added unit tests as well as linted using `make`. 

Screen Shot 2022-03-08 at 2 09 08 PM

**Checklist**

----

### DIFF
--- a/packages/cdk/lib/stacks/core-stack.ts
+++ b/packages/cdk/lib/stacks/core-stack.ts
@@ -81,8 +81,6 @@ export class CoreStack extends Stack {
         "idempotency-key": props.idempotencyKey,
       },
     });
-
-    console.log("context passed in App :point_right:", this.node.tryGetContext("fromApp"));
     new CfnOutput(this, VPC_PARAMETER_ID, { value: this.vpc.vpcId });
 
     const asset = new Asset(this, "WesAdapter", {
--- a/packages/cdk/lib/stacks/core-stack.ts
+++ b/packages/cdk/lib/stacks/core-stack.ts
@@ -4,7 +4,7 @@ import { StringParameter, IParameter } from "aws-cdk-lib/aws-ssm";
 import { GatewayVpcEndpointAwsService, InterfaceVpcEndpointService, IVpc, Vpc } from "aws-cdk-lib/aws-ec2";
 import { Bucket, BucketEncryption, IBucket } from "aws-cdk-lib/aws-s3";
 import { Construct } from "constructs";
-import { PRODUCT_NAME, APP_NAME, VPC_PARAMETER_NAME, WES_KEY_PARAMETER_NAME, WES_BUCKET_NAME } from "../constants";
+import { PRODUCT_NAME, APP_NAME, VPC_PARAMETER_NAME, WES_KEY_PARAMETER_NAME, WES_BUCKET_NAME, VPC_PARAMETER_ID } from "../constants";
 import { BucketDeployment, Source } from "aws-cdk-lib/aws-s3-deployment";
 import * as path from "path";
 import { homedir } from "os";
@@ -81,6 +81,9 @@ export class CoreStack extends Stack {
         "idempotency-key": props.idempotencyKey,
       },
     });
+
+    console.log("context passed in App :point_right:", this.node.tryGetContext("fromApp"));
+    new CfnOutput(this, VPC_PARAMETER_ID, { value: this.vpc.vpcId });
 
     const asset = new Asset(this, "WesAdapter", {
       path: path.join(homedir(), ".agc", "wes", "wes_adapter.zip"),
--- a/packages/cli/internal/pkg/cli/account_activate.go
+++ b/packages/cli/internal/pkg/cli/account_activate.go
@@ -188,7 +188,7 @@ func displayProgress(progressStream cdk.ProgressStream, displayMsg string) error
 			}
 			lastEvent = event
 		}
-		log.Info().Msg(strings.Join(lastEvent.Outputs, "\n"))
+		log.Debug().Msg(strings.Join(lastEvent.Outputs, "\n"))
 	} else {
 		return progressStream.DisplayProgress(displayMsg)
 	}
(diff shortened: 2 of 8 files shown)


packages/cdk/lib/stacks/core-stack.ts
```diff
@@ -82,6 +82,9 @@ export class CoreStack extends Stack {
       },
     });
 
+    console.log("context passed in App :point_right:", this.node.tryGetContext("fromApp"));
```
tneely, line 85: Can you remove this log statement?

packages/cli/internal/pkg/cli/account_activate.go
```diff
@@ -33,10 +35,11 @@ A new VPC will be created if not specified.`
...
 otherwise the parsing will fail.`
-	cdkCoreDir   = ".agc/cdk/apps/core"
-	bucketPrefix = "agc"
-	activateKey  = "activate"
-	bootstrapKey = "bootstrap"
+	cdkCoreDir    = ".agc/cdk/apps/core"
+	bucketPrefix  = "agc"
+	activateKey   = "activate"
+	bootstrapKey  = "bootstrap"
+	coreStackName = "Agc-Core"
```
tneely, line 42: Can you move this to [awsresources.go](https://github.com/aws/amazon-genomics-cli/blob/main/packages/cli/internal/pkg/cli/awsresources/awsresources.go) as a `RenderCoreStackName()` function?

packages/cli/internal/pkg/cli/account_activate.go
```diff
@@ -61,62 +65,97 @@ func newAccountActivateOpts(vars accountActivateVars) (*accountActivateOpts, err
...
 	}
 
 	environmentVars := []string{
-		fmt.Sprintf("%s=%s", constants.AgcBucketNameEnvKey, o.bucketName),
-		fmt.Sprintf("%s=%t", constants.CreateBucketEnvKey, !exists),
-		fmt.Sprintf("%s=%s", constants.AgcVersionEnvKey, version.Version),
-	}
-	if o.vpcId != "" {
-		environmentVars = append(environmentVars, fmt.Sprintf("%s=%s", constants.VpcIdEnvKey, o.vpcId))
+		fmt.Sprintf("AGC_BUCKET_NAME=%s", o.bucketName),
```
tneely, line 118: Why are you no longer using the constants?

packages/cli/internal/pkg/cli/account_activate.go
```diff
@@ -61,62 +65,97 @@ func newAccountActivateOpts(vars accountActivateVars) (*accountActivateOpts, err
...
-	account, err := o.stsClient.GetAccount()
-	if err != nil {
-		return "", err
+func (o accountActivateOpts) getVpcId() (string, error) {
+	if o.vpcId != "" {
+		return o.vpcId, nil
+	} else {
+		stackOutputs, err := o.cfnClient.GetStackOutputs(coreStackName)
+		if err != nil {
+			if err == cfn.StackDoesNotExistError {
```
tneely, line 147: Go error handling can be weird. I think what you have works but can you use `errors.Is(err, cfn.StackDoesNotExistError)` instead?

packages/cdk/lib/stacks/core-stack.ts
OhCloud: np.

packages/cli/internal/pkg/cli/account_activate.go
```diff
@@ -61,62 +65,97 @@ func newAccountActivateOpts(vars accountActivateVars) (*accountActivateOpts, err
...
+
+func (o accountActivateOpts) generateDefaultBucket() (string, error) {
+	account, err := o.stsClient.GetAccount()
+	if err != nil {
+		return "", err
+	}
+	return generateBucketName(account, o.region), nil
+}
+
+func (o accountActivateOpts) generateEnvVars() ([]string, error) {
```
a-li, line 103: I understand that you've created this function as `Execute()` is getting too big. Although you're catching the err on line 97, the err handling flow has changed and the program now wait until all these checks are complete as opposed to original fail-fast which is a better mechanism in my opinion. 

Let me know if I'm missing something.

a-li:
@OhCloud Consider updating this with current behavior: https://github.com/aws/amazon-genomics-cli/blob/main/site/content/en/docs/Concepts/accounts.md?plain=1#L42

packages/cli/internal/pkg/cli/account_activate.go
OhCloud: this was an odd import error. constants have now been restored.

packages/cli/internal/pkg/cli/account_activate.go
elliot-smith: Hey Angela,
If you look through the function, we exit out of the fail whenever the function returns an error so logically the code is exactly the same but it is easier to read. 
I.e. the error from line 97 is passed to line 108 which is then passed to line 79. The original flow would have been 97 to line 79.

a-li: Thanks Elliot for the detail explanation. I read it more closely now. It is doing I was intended. No change is required here.

elliot-smith (dismissed):

a-li (dismissed):

packages/cli/internal/pkg/cli/account_activate.go
```diff
@@ -147,6 +188,7 @@ func displayProgress(progressStream cdk.ProgressStream, displayMsg string) error
 			}
 			lastEvent = event
 		}
+		log.Info().Msg(strings.Join(lastEvent.Outputs, "\n"))
```
tneely, line 191: I just want to verify this is an intended change and not some debugging log.

OhCloud: running `account activate -v` will return a neat set of outputs including the TableName, VpcId, and Bucket. 
Screen Shot 2022-03-08 at 2 09 08 PM

tneely (approved):

PR #374, opened 2022-03-15
fix: Support unquoted int exit code

adamnovak:
Issue #, if available: #314

**Description of Changes**

This adds some custom unmarshalling logic to the Log object that should let it support both quoted and unquoted integral exit codes in JSON input.

I haven't validated my changes yet.

**Checklist**

----


adamnovak:
@nbraid, I'm going to fix #314 so I can use unmodified Toil with AGC.

Do you have any idea what's breaking the linter here? It seems like it can't even run.

nbraid:
It looks like there's a difference in the go.mod file? I'm not sure what could be causing that if you haven't modified the go.mod file.

```
Run git diff --exit-code
diff --git a/packages/cli/go.mod b/packages/cli/go.mod
index bf3f117..99df139 100644
--- a/packages/cli/go.mod
+++ b/packages/cli/go.mod
@@ -35,7 +35,7 @@ require (
 	github.com/stretchr/testify v1.7.0
 	github.com/xeipuuv/gojsonschema v1.2.0
 	golang.org/x/net v0.0.0-20211015210444-4f30a5c0130f
-	golang.org/x/tools v0.1.9 // indirect
+	golang.org/x/tools v0.1.10 // indirect
 	gopkg.in/check.v1 v1.0.0-20190902080502-41f04d3bba15 // indirect
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b
 )
```

nbraid:
Sounds like this is because we run `go mod vendor` in the check, which will update to the latest version. You might need to run `go mod vendor` from the `package/cli` directory so your commit matches what our built version will download.

adamnovak:
When I check this out and run `go mod vendor` in `packages/cli`, it makes no changes to `go.mod` in that directory. Running `make` doesn't make any changes either.

I manually bumped the `tools` version; maybe that will help?

adamnovak:
I've pulled the Makefile down into the included WES client module, and given its test and formatting steps (which is why I'm reformatting all its files).

And I added some unit tests for deserializing example log objects of the different shapes we want to support.

So I think this is now ready for review! Hopefully CI likes it.

adamnovak:
@nbraid This now has tests and passes CI.

elliot-smith:
Hi Adam,
We currently have a PR that is about to be raised that reverts the exit code back to an integer (including updating the relevant AGC code) which I believe will suit your use case better. Are you okay if we merge that change in over this change? It should be raised within the next couple of days.

adamnovak:
@elliot-smith That should be fine for us; Toil always sends an integer.

elliot-smith:
Closing this PR in favor of this one (as discussed above): https://github.com/aws/amazon-genomics-cli/pull/380

Issue #357, opened 2022-03-05
nextflow wes adapter does not return more than 100 tasks


wleepang:
[//]: #  (Please do not include your AWS account or credentials as part of the bug request.)

**Describe the Bug**

[//]: #  (Please describe the behavior that you are seeing with as many details as possible including logs from AGC, CloudWatch snippets and CloudFormation errors if relevant.)

The WES adapter and endpoint for contexts that use `nextflow` as an engine only returns details for up to 100 tasks in the WES GetRunLog response. Subsequently, `agc logs workflow` output only shows up to 100 tasks. 

**Steps to Reproduce**

[//]: #  (Include a list of steps that can be used to reproduce the bug.)

I created a workflow that generates 1000 tasks in both wdl and nextflow.

**wdl workflow**
```wdl
version 1.0

workflow TestSimple1000 {
    Array[Int] numbers = range(1000)
    scatter ( number in numbers ) {
        call PrintNumber {
            input:
                number = number
        }
    }
}

task PrintNumber {
    input {
        Int number
    }

    command {
        echo ~{number}
    }

    runtime {
        docker: "public.ecr.aws/lts/ubuntu:latest"
        memory: "512 MB"
    }
}
```

**nextflow workflow**
```nextflow
nextflow.enable.dsl=2

process PrintNumber {
    container "public.ecr.aws/lts/ubuntu:latest"
    memory "512 MB"
    input:
        val number

    script:
        println "${number}"
        """
        echo "${number}"
        """
}

workflow {
    range = 0..999
    numbers = Channel.fromList(range.toList())
    PrintNumber(numbers)
}
```

I created an agc project for these workflows with the following file structure:

* project/
  * wdl/
    * test-simple-workflow/
      * main.wdl
      * MANIFEST.json
  * nextflow/
    * test-simple-workflow/
      * main.nf
      * MANIFEST.json
  * agc-project.yaml

The project configuration is:

```yaml
---
name: Test1000
schemaVersion: 1
workflows:
  test-simple-1000-wdl:
    type:
      language: wdl
      version: 1.0
    sourceURL: ./wdl/test-simple-1000
  test-simple-1000-nextflow:
    type:
      language: nextflow
      version: dsl2
    sourceURL: ./nextflow/test-simple-1000
contexts:
  onDemandCtxWdl:
    engines:
      - type: wdl
        engine: cromwell
  onDemandCtxNextflow:
    engines:
      - type: nextflow
        engine: nextflow
...
```

[...]

PR #377, opened 2022-03-18
fix: nextflow wes adapter can retrieve more than 100 task logs

wleepang:
Issue #, if available: #357

**Description of Changes**

Updated the WES Adapter for Nextflow to be able to retrieve job details for more than 100 tasks.

* Increased the CloudWatchLogs query results limit to maximum (10,000)
* Updated BatchAdapter `describe_jobs` method to retrieve more than 100 job descriptions.

Using an installation of AGC:

* unzipped `~/.agc/wes/wes_adapter.zip`
* modified code in `wes_adpater/amazon_genomics/
* recreated `~/.agc/wes/wes_adapter.zip`
* ran `agc account activate` to redeploy assets for CDK
* ran `agc context deploy` to redeploy affected nextflow context
* ran test workflow "test-simple-1000" with 1000 jobs
* ran `agc logs workflow test-simple-1000 -r <run-id>` to retrieve logs
* counted lines of tasks to verify more than 100 returned:

```
$ agc logs workflow test-simple-1000 -r 1fe54a99-5431-493c-b7f6-f0af6cffd8fa | wc -l
2022-03-18T04:02:30Z 𝒊  Showing the logs for 'test-simple-1000'
1005
```

**Checklist**

----


wleepang:
A note on this PR:

This is a relatively simple fix and **does not** address the edge case of a workflow that has more than 10,000 sub-tasks. In the case of human genomics, it is possible for a workflow to have a scatter step that submits >10,000 jobs within a short period of time - e.g. processing intervals for all genes in the genome.

To cover that case one needs to update the `query_logs_for_job` method in `NextflowWESAdapter` to handle multiple CloudWatchLogs queries but also be cognizant of the concurrent query limit of 10.

packages/wes_adapter/amazon_genomics/wes/adapters/NextflowWESAdapter.py
```diff
@@ -136,7 +136,7 @@ def query_logs_for_job(self, job_details: JobDetailTypeDef, query: str):
             endTime=end_time or int(time.time()),
             queryString=query,
             # TODO: handle pagination? GetRunLog doesn't seem to support it...
-            limit=100,
+            limit=10_000,
```
markjschreiber, line 139: Does this need to be 10000 if chunks are 100? Alternatively could chunk size be limit?

markjschreiber (commented):
Looks good. One query about chunk size.

markjschreiber:
The build issue seems to be with formatting. If you run `make` it should fix it

wleepang:
Removed the wait from the `describe_jobs` method in `BatchAdapter`. Upon testing with a more realistic workflow with slightly more than 1000 jobs, the time to retrieve all job descriptions was over the timeout limit for API Gateway and creating 504 errors.

wleepang:
Increased headroom on WES adapter lambda by doing the following:

1. Added multithreaded concurrency for the Batch DescribeJobs operation
2. Increased Lambda Function memory to 256MB

This reduces the API latency by roughly half - from 21s to about 10s

packages/wes_adapter/amazon_genomics/wes/adapters/NextflowWESAdapter.py
elliot-smith, line 139: Is this meant to be an underscore?

packages/wes_adapter/amazon_genomics/wes/adapters/BatchAdapter.py
```diff
@@ -180,7 +183,25 @@ def describe_job(self, job_id: str) -> Optional[JobDetailTypeDef]:
     def describe_jobs(self, job_ids: typing.List[str]) -> typing.List[JobDetailTypeDef]:
         if not job_ids:
             return []
-        return self.aws_batch.describe_jobs(jobs=job_ids)["jobs"]
+
+        jobs = []
+        job_ids_sets = chunks(job_ids, 100)
+        with ThreadPoolExecutor(max_workers=10) as executor:
```
elliot-smith, line 189: Can you please add another unit test that tests the branching logic here? We are only testing a single retrieval request atm.

wleepang:
One additional minor fix - start and end times are in utc, if end_time is not available, use the `datetime.utcnow()` value instead of `time.time()`. The latter pulls from the host compute's locale.

packages/wes_adapter/amazon_genomics/wes/adapters/NextflowWESAdapter.py
wleepang: Yes. The underscore is optional but a good idea for readability of large numbers:
https://peps.python.org/pep-0515/

packages/wes_adapter/amazon_genomics/wes/adapters/NextflowWESAdapter.py
```diff
@@ -133,10 +133,10 @@ def query_logs_for_job(self, job_details: JobDetailTypeDef, query: str):
         query_id = self.aws_logs.start_query(
             logGroupName=self.engine_log_group,
             startTime=start_time,
-            endTime=end_time or int(time.time()),
+            endTime=end_time or int(datetime.utcnow().timestamp()),
```
markjschreiber, line 136: After much head scratching I discovered that `datetime.utcnow().timestamp()` gives you seconds but `startTime` is in milliseconds. You want something like: `int(math.ceil(datetime.utcnow().timestamp()) * 1000)`. `math.ceil` is needed because of a potential race condition on the timestamp in seconds being less than the timestamp in milliseconds which can contain fractions of seconds.

markjschreiber (changes requested):

packages/wes_adapter/amazon_genomics/wes/adapters/BatchAdapter.py
wleepang: Done. The associated test passes. However there are three unit tests in BatchAdapter that do not pass. Switching to `main`, they did't pass before my changes.

markjschreiber: Can you remove them and file an issue to fix or replace them noting the commit ID before they were removed?

packages/wes_adapter/amazon_genomics/wes/adapters/BatchAdapter.py
```diff
@@ -180,7 +183,25 @@ def describe_job(self, job_id: str) -> Optional[JobDetailTypeDef]:
...
+                ): job_ids_set
+                for job_ids_set in job_ids_sets
+            }
+            for future in as_completed(future_jobs):
+                job_ids_set = future_jobs[future]
+                try:
+                    response = future.result()
+                    jobs += response["jobs"]
+                except Exception as e:
+                    print(f"error retrieving jobs: {e}")
```
markjschreiber, line 202: Can `self.logger` be used here for consistency with other classes?

packages/wes_adapter/amazon_genomics/wes/adapters/BatchAdapter.py
```diff
@@ -180,7 +183,25 @@ def describe_job(self, job_id: str) -> Optional[JobDetailTypeDef]:
...
+            future_jobs = {
+                executor.submit(
+                    self.aws_batch.describe_jobs, jobs=job_ids_set
+                ): job_ids_set
+                for job_ids_set in job_ids_sets
+            }
+            for future in as_completed(future_jobs):
+                job_ids_set = future_jobs[future]
+                try:
+                    response = future.result()
```
markjschreiber, line 199: Does `future.result()` offer a time out? Without any timeout the adapter could hang.

markjschreiber (commented):
Some questions and suggestions

packages/wes_adapter/amazon_genomics/wes/adapters/BatchAdapter.py
wleepang: Yes

packages/wes_adapter/amazon_genomics/wes/adapters/BatchAdapter.py
wleepang: Yes. Can set this to 5s

markjschreiber (dismissed):

packages/wes_adapter/amazon_genomics/wes/adapters/BatchAdapter.py
wleepang: I've kept the unit tests as is and created issue #386

wleepang:
Added one additional unit test that tests the number of jobIds returned by `BatchAdapter.describe_jobs`

markjschreiber (approved):

elliot-smith (approved):

PR #379, opened 2022-03-22
feat: Engine logs for specific workflow runs

markjschreiber:
Issue #, if available:

**Description of Changes**

Workflows that run in engines that run as "head" processes produce distinct log streams in their log group for each run. To be useful, the `agc logs engine` command needs to be able to specify the ID of the run of interest.

A flag `-r` or `--run-id` has been added.
* If this flag is NOT set the user is warned that it should be set and that not including it will be required in the future, all streams from the log group will be displayed. This is to maintain backward compatibility.
* If the flag is set with Cromwell then an error is emitted indicating this flag is currently not supported for cromwell (adding support, if even possible, is a future task).

### Edge Cases
1. Log stream name is not known to WES API but run-id is valid. Workflow engine is not started: User is warned that the stream is not present and given the current workflow state.
1. Log stream not available from CloudWatch but log stream name is known to WES API. Workflow engine process started but CloudWatch is not yet fully consistent. User is warned that the logging process has started but the logstream is not yet available and they should wait and try again 

Added and updated unit tests
* `make init`
* `make`

For each "head process" engine:
 * `agc deploy context -c <context-name>`
 * `RUN_ID=agc workflow run -c <context-name> hello`
 * `for i in {0..30}; do agc logs engine -c <context-name> -r $RUN_ID && sleep 3`; done`

Ensure no errors during workflow initialization and logs appear.

To test tailing use the same procedure but run : 
* `for i in {0..30}; do agc logs engine -c <context-name> -r $RUN_ID -t && sleep 3`; done`

For Cromwell confirm that engine logs and log tailing are the same as for previous versions of AGC.

**Checklist**

----

### DIFF
--- a/packages/cli/internal/pkg/cli/context/info.go
+++ b/packages/cli/internal/pkg/cli/context/info.go
@@ -1,10 +1,13 @@
 package context
 
 import (
-	"reflect"
-
 	"github.com/aws/amazon-genomics-cli/internal/pkg/cli/spec"
+	"reflect"
+	"strings"
 )
+
+var serverEngines = map[string]bool{"cromwell": true}
+var headEngines = map[string]bool{"nextflow": true, "miniwdl": true, "snakemake": true}
 
 type Summary struct {
 	Name          string
@@ -16,6 +19,18 @@ type Summary struct {
 
 func (i Summary) IsEmpty() bool {
 	return reflect.ValueOf(i).IsZero()
+}
+
+//IsServerProcessEngine Does the workflow engine run as a server process? A server process has a potential one to many
+// mapping with workflow runs.
+func (i *Summary) IsServerProcessEngine() bool {
+	return serverEngines[strings.ToLower(strings.TrimSpace(i.Engines[0].Engine))]
+}
+
+//IsHeadProcessEngine Does the workflow engine run as a head process? A head process has one to one
+// mapping with workflow runs. Processes are not reused.
+func (i *Summary) IsHeadProcessEngine() bool {
+	return headEngines[strings.ToLower(strings.TrimSpace(i.Engines[0].Engine))]
 }
 
 type Detail struct {
--- a/packages/cli/internal/pkg/cli/context/info.go
+++ b/packages/cli/internal/pkg/cli/context/info.go
@@ -7,7 +7,7 @@ import (
 	"github.com/aws/amazon-genomics-cli/internal/pkg/constants"
 )
 
-var headEngines = map[string]bool{constants.NEXTFLOW: true, constants.MINIWDL: true, constants.SNAKEMAKE: true}
+var serverProcessEngines = map[string]bool{constants.CROMWELL: true}
 
 type Summary struct {
 	Name          string
@@ -21,10 +21,10 @@ func (s Summary) IsEmpty() bool {
 	return reflect.ValueOf(s).IsZero()
 }
 
-//IsHeadProcessEngine Does the workflow engine run as a head process? A head process has one to one
-// mapping with workflow runs. Processes are not reused.
-func (s *Summary) IsHeadProcessEngine() bool {
-	return headEngines[s.Engines[0].Engine]
+//IsServerProcessEngine Does the workflow engine run as a server process? A server process engine has one to many
+// mapping with workflow runs. The engine can be used to run multiple workflows and the process is re-used and long running.
+func (s *Summary) IsServerProcessEngine() bool {
+	return serverProcessEngines[s.Engines[0].Engine]
 }
 
 type Detail struct {
--- a/packages/cli/internal/pkg/cli/context/info.go
+++ b/packages/cli/internal/pkg/cli/context/info.go
@@ -1,9 +1,10 @@
 package context
 
 import (
-	"github.com/aws/amazon-genomics-cli/internal/pkg/cli/spec"
 	"reflect"
 	"strings"
+
+	"github.com/aws/amazon-genomics-cli/internal/pkg/cli/spec"
 )
 
 var serverEngines = map[string]bool{"cromwell": true}
--- a/packages/cli/internal/pkg/cli/context/info.go
+++ b/packages/cli/internal/pkg/cli/context/info.go
@@ -2,13 +2,12 @@ package context
 
 import (
 	"reflect"
-	"strings"
 
 	"github.com/aws/amazon-genomics-cli/internal/pkg/cli/spec"
+	"github.com/aws/amazon-genomics-cli/internal/pkg/constants"
 )
 
-var serverEngines = map[string]bool{"cromwell": true}
-var headEngines = map[string]bool{"nextflow": true, "miniwdl": true, "snakemake": true}
+var headEngines = map[string]bool{constants.NEXTFLOW: true, constants.MINIWDL: true, constants.SNAKEMAKE: true}
 
 type Summary struct {
 	Name          string
@@ -18,20 +17,14 @@ type Summary struct {
 	Engines       []spec.Engine
 }
 
-func (i Summary) IsEmpty() bool {
-	return reflect.ValueOf(i).IsZero()
-}
-
-//IsServerProcessEngine Does the workflow engine run as a server process? A server process has a potential one to many
-// mapping with workflow runs.
-func (i *Summary) IsServerProcessEngine() bool {
-	return serverEngines[strings.ToLower(strings.TrimSpace(i.Engines[0].Engine))]
+func (s Summary) IsEmpty() bool {
+	return reflect.ValueOf(s).IsZero()
 }
 
 //IsHeadProcessEngine Does the workflow engine run as a head process? A head process has one to one
 // mapping with workflow runs. Processes are not reused.
-func (i *Summary) IsHeadProcessEngine() bool {
-	return headEngines[strings.ToLower(strings.TrimSpace(i.Engines[0].Engine))]
+func (s *Summary) IsHeadProcessEngine() bool {
+	return headEngines[s.Engines[0].Engine]
 }
 
 type Detail struct {
@@ -52,6 +45,6 @@ type Instance struct {
 	IsDefinedInProjectFile bool
 }
 
-func (i Detail) IsEmpty() bool {
-	return reflect.ValueOf(i).IsZero()
+func (d Detail) IsEmpty() bool {
+	return reflect.ValueOf(d).IsZero()
 }
--- a/packages/cli/internal/pkg/cli/context/info_test.go
+++ b/packages/cli/internal/pkg/cli/context/info_test.go
@@ -28,32 +28,41 @@ func TestDetail_IsNotEmpty(t *testing.T) {
 	assert.False(t, detail.IsEmpty())
 }
 
-func TestSummary_IsHeadProcessEngine(t *testing.T) {
-	type testScenario struct {
+func TestSummary_IsServerProcessEngine(t *testing.T) {
+	tests := map[string]struct {
 		engineName string
 		expect     bool
+	}{
+		"otherIsNotAServer": {
+			engineName: "other",
+			expect:     false,
+		},
+
+		"cromwellIsAServer": {
+			engineName: constants.CROMWELL,
+			expect:     true,
+		},
+
+		"snakeMakeIsNotAServer": {
+			engineName: constants.SNAKEMAKE,
+			expect:     false,
+		},
+		"nextFlowIsNotAServer": {
+			engineName: constants.NEXTFLOW,
+			expect:     false,
+		},
+		"miniwdlIsNotAServer": {
+			engineName: constants.MINIWDL,
+			expect:     false,
+		},
 	}
 
-	scenarios := []testScenario{{
-		engineName: "other",
-		expect:     false,
-	}, {
-		engineName: constants.CROMWELL,
-		expect:     false,
-	}, {
-		engineName: constants.SNAKEMAKE,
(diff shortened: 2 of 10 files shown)


packages/cli/internal/pkg/cli/context/info.go
```diff
@@ -2,10 +2,14 @@ package context
 
 import (
 	"reflect"
+	"strings"
 
 	"github.com/aws/amazon-genomics-cli/internal/pkg/cli/spec"
 )
 
+var serverEngines = map[string]bool{"cromwell": true}
+var headEngines = map[string]bool{"nextflow": true, "miniwdl": true, "snakemake": true}
```
IllyaYalovyy, line 11: I would prefer to not have engine names as strings. We may want to have them as constants in some shared package. Here is an example of desired usage:

```go
var serverBasedEngine = map[engine.Name]bool{engine.NEXTFLOW: true}
```

packages/cli/internal/pkg/cli/context/info.go
```diff
@@ -18,6 +22,18 @@ func (i Summary) IsEmpty() bool {
 	return reflect.ValueOf(i).IsZero()
 }
 
+//IsServerProcessEngine Does the workflow engine run as a server process? A server process has a potential one to many
+// mapping with workflow runs.
+func (i *Summary) IsServerProcessEngine() bool {
+	return serverEngines[strings.ToLower(strings.TrimSpace(i.Engines[0].Engine))]
```
IllyaYalovyy, line 28: I think in our case it can be safely simplified to just:

```go
func (s *Summary) IsServerProcessEngine() bool {
   return s.Engines[0].Engine == engine.NEXTFLOW
}
```

packages/cli/internal/pkg/cli/context/info.go
```diff
@@ -18,6 +22,18 @@ func (i Summary) IsEmpty() bool {
 	return reflect.ValueOf(i).IsZero()
 }
 
+//IsServerProcessEngine Does the workflow engine run as a server process? A server process has a potential one to many
+// mapping with workflow runs.
+func (i *Summary) IsServerProcessEngine() bool {
```
IllyaYalovyy, line 27: `i` is usually used as an index in loops. I would recommend to rename it to `s`.

packages/cli/internal/pkg/cli/context/info.go
```diff
@@ -18,6 +22,18 @@ func (i Summary) IsEmpty() bool {
...
+	return serverEngines[strings.ToLower(strings.TrimSpace(i.Engines[0].Engine))]
+}
+
+//IsHeadProcessEngine Does the workflow engine run as a head process? A head process has one to one
+// mapping with workflow runs. Processes are not reused.
+func (i *Summary) IsHeadProcessEngine() bool {
+	return headEngines[strings.ToLower(strings.TrimSpace(i.Engines[0].Engine))]
```
IllyaYalovyy, line 34: If we expect Engine name to be in a wrong format (padding spaces, mixed case), I think it will be better to normalize it on ingestion, so we can avoid repeating this boilerplate code in all places. There is a potential risk to create inconsistent behavior and inflate number of unit tests. Instead of testing every method that touches Engine for trim and lowerCase, we can test one.

packages/cli/internal/pkg/cli/context/info_test.go
```diff
@@ -25,3 +29,35 @@ func TestDetail_IsNotEmpty(t *testing.T) {
 	detail := Detail{WesUrl: "amazon.com"}
 	assert.False(t, detail.IsEmpty())
 }
+
+func TestSummary_IsHeadProcessEngine_HeadEnginesShouldReturnTrue(t *testing.T) {
+	for _, engineName := range testHeadEngineNames {
```
IllyaYalovyy, line 34: Please use a parameterized tests. It is better to avoid testing multiple cases in one test.
For more details please read:

https://elliotchance.medium.com/go-data-driven-or-parameterized-tests-cd2b62b21c8a

packages/cli/internal/pkg/cli/logs_engine.go
```diff
@@ -45,10 +72,29 @@ func (o *logsEngineOpts) Execute() error {
 	}
 
 	logGroupName := contextInfo.EngineLogGroupName
-	if o.tail {
-		err = o.followLogGroup(logGroupName)
+	log.Debug().Msgf("Engine log group name: '%s'", logGroupName)
+
+	if o.workflowRunId != "" {
```
IllyaYalovyy, line 77: This method is getting a bit too big. Could you please simplify it by extracting independent flows into separate methods. I would prefer it to look like this:

```go
if o.workflowRunId != "" {
  return o.executeGetWholeEngineLog()
} else {
  return o.executeGetEngineLogForRunId()
}
```

packages/cli/internal/pkg/cli/logs_engine.go
```diff
@@ -45,10 +72,29 @@ func (o *logsEngineOpts) Execute() error {
...
+		log.Debug().Msgf("Getting log stream for workflow run '%s'", o.workflowRunId)
+
+		workflowRunLog, err := o.workflowManager.GetEngineLogByRunId(o.workflowRunId)
+		if err != nil {
+			return err
+		}
+		log.Debug().Msgf("Stream '%s' for log group '%s' contains StdOut for run '%s'", workflowRunLog.StdOut, logGroupName, workflowRunLog.WorkflowRunId)
+
+		if o.tail {
+			err = o.followLogStreams(logGroupName, workflowRunLog.StdOut)
```
IllyaYalovyy, line 87:
```go
return o.followLogStreams(logGroupName, workflowRunLog.StdOut)
```

packages/cli/internal/pkg/cli/logs_engine.go
```diff
@@ -45,10 +72,29 @@ func (o *logsEngineOpts) Execute() error {
...
+		} else {
+			err = o.displayLogStreams(logGroupName, o.startTime, o.endTime, o.filter, workflowRunLog.StdOut)
```
IllyaYalovyy, line 89: ditto

packages/cli/internal/pkg/cli/logs_engine.go
```diff
@@ -45,10 +72,29 @@ func (o *logsEngineOpts) Execute() error {
...
+		}
+
 	} else {
-		err = o.displayLogGroup(logGroupName, o.startTime, o.endTime, o.filter)
+		if o.tail {
+			err = o.followLogGroup(logGroupName)
```
IllyaYalovyy, line 94: ditto

packages/cli/internal/pkg/cli/logs_engine.go
```diff
@@ -45,10 +72,29 @@ func (o *logsEngineOpts) Execute() error {
...
+	log.Debug().Msgf("Engine log group name: '%s'", logGroupName)
+
+	if o.workflowRunId != "" {
+		log.Debug().Msgf("Getting log stream for workflow run '%s'", o.workflowRunId)
+
+		workflowRunLog, err := o.workflowManager.GetEngineLogByRunId(o.workflowRunId)
+		if err != nil {
+			return err
+		}
+		log.Debug().Msgf("Stream '%s' for log group '%s' contains StdOut for run '%s'", workflowRunLog.StdOut, logGroupName, workflowRunLog.WorkflowRunId)
```
IllyaYalovyy, line 84: We can improve readability if we do:

```go
logStreamNames := workflowRunLog.StdOut
```

packages/cli/internal/pkg/cli/context/info.go
IllyaYalovyy, line 27: Where is this function being used?

packages/cli/internal/pkg/cli/logs_engine.go
```diff
@@ -45,10 +72,29 @@ func (o *logsEngineOpts) Execute() error {
...
 
 	logGroupName := contextInfo.EngineLogGroupName
-	if o.tail {
-		err = o.followLogGroup(logGroupName)
+	log.Debug().Msgf("Engine log group name: '%s'", logGroupName)
+
+	if o.workflowRunId != "" {
+		log.Debug().Msgf("Getting log stream for workflow run '%s'", o.workflowRunId)
+
+		workflowRunLog, err := o.workflowManager.GetEngineLogByRunId(o.workflowRunId)
```
IllyaYalovyy, line 80: Some engines can return an empty list of coudwatch streams. In this case the downstream code will dump complete engine logs. We should handle this edge case.

IllyaYalovyy (changes requested):

packages/cli/internal/pkg/cli/context/info.go
```diff
@@ -4,8 +4,11 @@ import (
 	"reflect"
 
 	"github.com/aws/amazon-genomics-cli/internal/pkg/cli/spec"
+	"github.com/aws/amazon-genomics-cli/internal/pkg/constants"
 )
 
+var headEngines = map[string]bool{constants.NEXTFLOW: true, constants.MINIWDL: true, constants.SNAKEMAKE: true}
```
IllyaYalovyy, line 10: From maintainability point of view, all future engines most likely will be "HeadEngine". It means every time we add a new engine we will have to update this line (easy to forget). It would be more reliable if we do the opposite. We can store only one (legacy) engine which is NOT a Head Engine. This way our implementation will be future-proof.

packages/cli/internal/pkg/cli/context/info_test.go
```diff
@@ -25,3 +27,33 @@ func TestDetail_IsNotEmpty(t *testing.T) {
 	detail := Detail{WesUrl: "amazon.com"}
 	assert.False(t, detail.IsEmpty())
 }
+
+func TestSummary_IsHeadProcessEngine(t *testing.T) {
```
IllyaYalovyy, line 31: Could you please make your test consistent with other tests?

Here is a good example that I would suggest to follow:
https://github.com/aws/amazon-genomics-cli/blob/2a8fab00ceedfd72321f5466183edf92f4c4d0fa/packages/cli/internal/pkg/cli/logs_workflow_test.go#L65

packages/cli/internal/pkg/cli/logs_engine.go
```diff
@@ -45,24 +79,65 @@ func (o *logsEngineOpts) Execute() error {
 	}
 
 	logGroupName := contextInfo.EngineLogGroupName
-	if o.tail {
-		err = o.followLogGroup(logGroupName)
+	log.Debug().Msgf("Engine log group name: '%s'", logGroupName)
+
+	if o.workflowRunId == "" {
+		err = executeGetEngineLogForWholeGroup(o, logGroupName)
```
IllyaYalovyy, line 85: can be simplified
```go
if o.workflowRunId == "" {
  return executeGetEngineLogForWholeGroup(o, logGroupName)
}
return executeGetEngineLogForRunId(o, logGroupName)
```

packages/cli/internal/pkg/cli/logs_engine.go
```diff
@@ -45,24 +79,65 @@ func (o *logsEngineOpts) Execute() error {
...
+		return o.displayLogStreams(logGroupName, o.startTime, o.endTime, o.filter, logStreamName)
+	}
+
+	workflowStatus := workflowRunLog.WorkflowStatus
+	log.Warn().Msgf("Cannot find an engine log stream for workflow run '%s', the current status of the run is: '%s', "+
+		"the log will not be available until after the workflow is RUNNING", o.workflowRunId, workflowStatus)
+	return nil
+}
+
+func streamNameFromRunLog(workflowRunLog workflow.EngineLog) string {
```
IllyaYalovyy, line 124: I think it should be a list of streams. It is possible we have stderr and stdout. I think assumption that we always have only one or another may result in sub-optimal user experience, when not all logs will be presented.

IllyaYalovyy (changes requested):

packages/cli/internal/pkg/cli/context/info.go
markjschreiber: Good idea. I'll keep it as a map because I think Toil will be a server engine as well but these are the exception rather than the rule.

IllyaYalovyy (commented):
Do you have to commit "packages/cdk/npm-shrinkwrap.json" ? It seems irrelevant to the topic of this CR.

packages/cli/internal/pkg/cli/logs_engine.go
```diff
@@ -35,6 +50,27 @@ func (o *logsEngineOpts) Validate() error {
...
+	if err != nil {
+		return err
+	}
+
+	summary := ctxMap[o.contextName]
+	engine := summary.Engines[0].Engine
+
+	if o.workflowRunId == "" {
+		if !summary.IsServerProcessEngine() {
+			log.Warn().Msgf("DEPRECATION WARNING!!")
```
IllyaYalovyy, line 63: Are we going to deprecate this? Full engine log has parts related to engine work itself (at least for Cromwell).

markjschreiber: For the "serverless" engines yes that's why the warning is inside the if statement.

markjschreiber:
> Do you have to commit "packages/cdk/npm-shrinkwrap.json" ? It seems irrelevant to the topic of this CR.

It's unclear. It was updated during a `make` build so I guess `npm` wanted to update it.

IllyaYalovyy (approved):

elliot-smith (approved):

Issue #205, opened 2021-12-02
miniwdl engine needs to be able to terminate running and queued tasks when a failure happens


markjschreiber:
[//]: #  (Please do not include your AWS account or credentials as part of the bug request.)

**Describe the Bug**

When a task fails the miniwdl head task will attempt to terminate all currently running tasks from the workflow.

**Steps to Reproduce**

Run a workflow with a scatter where one of the tasks returns with a non-zero return code.

**Relevant Logs**

example line

```
2021-12-02 18:57:46.475 wdl.w:BasicJointGenotyping3000SamplesChrOne.t:call-ImportGVCFs-1433 ERROR AccessDeniedException, User: arn:aws:sts::123456789012:assumed-role/Agc-Context-ScaledJointGe-miniwdlHeadBatchBatchRol-IO0I9REMOUDL/1040aebb4a584db1be4383b79cfa1cd8 is not authorized to perform: batch:TerminateJob on resource: arn:aws:batch:us-west-2: 123456789012:job/9bc7488f-e2ae-4d59-b6a5-37bdfa8be6d8
```

**Expected Behavior**

Head job's IAM policy should allow it to terminate any currently running tasks or queued tasks from the workflow that the head task is coordinating.

Ideally the policy should limit to *ONLY* being able to cancel/ terminate tasks being created by the head task.

**Actual Behavior**

* AccessDeniedException as see above
* Inspection of the IAM policy shows that it doesn't have the required permission

**Additional Context**

[//]: #  (Add any other context about the problem here. The minimum required information is below)

Operating System:
AGC Version: 1.1.2
Was AGC setup with a custom bucket: No
Was AGC setup with a custom VPC: No

PR #391, opened 2022-04-06
fix: ensure workflow engines can and do terminate child tasks when the engine task is stopped by AGC

markjschreiber:
Issue #, if available:
#205 

**Description of Changes**

* Updated IAM policies to ensure all engines have permissions to terminate child tasks
* Updated WES adapter IAM policies to ensure they have permission to terminate engine tasks
* Notified miniwdl-ext author of bug in handling `SIGTERM` and added fixed version to our miniwdl deployment
* Found that snakemake doesn't cancel child tasks when handling `SIGTERM` changed that behavior in our snakemake repo.

For all engines:

1. deploy context
2. run a workflow
3. wait until the workflow spawns child tasks that become queued or running in AWS Batch
4. stop workflow with `agc workflow stop`
5. observe that workflow tasks terminate with the expected status reason and for Batch engine tasks, ensure they also terminate.

**Checklist**

----

### DIFF
--- a/packages/cdk/lib/stacks/engines/miniwdl-engine-construct.ts
+++ b/packages/cdk/lib/stacks/engines/miniwdl-engine-construct.ts
@@ -1,8 +1,8 @@
-import { Stack, Aws } from "aws-cdk-lib";
+import { Aws, Stack } from "aws-cdk-lib";
 import { Bucket, IBucket } from "aws-cdk-lib/aws-s3";
 import { ApiProxy, Batch } from "../../constructs";
-import { EngineOutputs, EngineConstruct } from "./engine-construct";
-import { IRole, PolicyDocument, PolicyStatement, Role, ServicePrincipal, ManagedPolicy } from "aws-cdk-lib/aws-iam";
+import { EngineConstruct, EngineOutputs } from "./engine-construct";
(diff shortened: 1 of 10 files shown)


packages/cdk/lib/stacks/engines/miniwdl-engine-construct.ts
```diff
@@ -62,6 +70,11 @@ export class MiniwdlEngineConstruct extends EngineConstruct {
               actions: ["tag:GetResources"],
               resources: ["*"],
             }),
+            new PolicyStatement({
+              effect: Effect.ALLOW,
+              actions: ["batch:TerminateJob"],
+              resources: ["*"],
```
elliot-smith, line 76: I believe that we tag all of our instances with the following key value pair
```
"application-name": "agc"
```

Can we restrict the policy to just those instances (we should do this for the other engines as well)?

markjschreiber: This statement is attached to the WES adapter policy. Currently we don't tag our head engine fargate tasks so we can't restrict them. However, I will raise a new issue to do this because if we did then we could reduce access with something like. 

```
conditions: { StringEquals: { "aws:ResourceTag/application-name": "agc" } }
```

elliot-smith (approved):

tneely (approved):

PR #398, opened 2022-04-08
fix: updates cromwell exit_code

OhCloud:
Issue #, if available:

**Description of Changes**

Updates Cromwell to handle exit_code conversion (previously missed on #314)

Steps to reproduce [using cromwell]
```
account activate -c
context deploy -c myContext -v 
workflow run words-with-vowels -c myContext -v
logs workflow words-with-vowels -r <RunId>
```
Returns

Screen Shot 2022-04-08 at 12 07 04 PM

**Checklist**

----

### DIFF
--- a/packages/wes_adapter/amazon_genomics/wes/adapters/CromwellWESAdapter.py
+++ b/packages/wes_adapter/amazon_genomics/wes/adapters/CromwellWESAdapter.py
@@ -339,7 +339,7 @@ class CromwellWESAdapter(AbstractWESAdapter):  # inherit from ABC to enforce int
                     "end_time": task.get("end"),
                     "stdout": task.get("stdout"),
                     "stderr": task.get("stderr"),
-                    "exit_code": task.get("returnCode")
+                    "exit_code": task.get("returnCode"),
                 }
                 task_logs.append(log)
 
--- a/packages/wes_adapter/amazon_genomics/wes/adapters/CromwellWESAdapter.py
+++ b/packages/wes_adapter/amazon_genomics/wes/adapters/CromwellWESAdapter.py
@@ -339,11 +339,7 @@ class CromwellWESAdapter(AbstractWESAdapter):  # inherit from ABC to enforce int
                     "end_time": task.get("end"),
                     "stdout": task.get("stdout"),
                     "stderr": task.get("stderr"),
-                    "exit_code": (
-                        ""
-                        if task.get("returnCode") == None
-                        else str(task.get("returnCode"))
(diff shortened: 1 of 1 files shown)


elliot-smith (dismissed):

packages/wes_adapter/amazon_genomics/wes/adapters/CromwellWESAdapter.py
```diff
@@ -339,11 +339,7 @@ def _build_run_log_dict_(self, metadata, outputs):
                     "end_time": task.get("end"),
                     "stdout": task.get("stdout"),
                     "stderr": task.get("stderr"),
-                    "exit_code": (
-                        ""
-                        if task.get("returnCode") == None
-                        else str(task.get("returnCode"))
-                    ),
+                    "exit_code": task.get("returnCode")
```
markjschreiber: Do we expect `task.get("returnCode")` to be a number? If so you might be able to use `int(task.get("returnCode"))` to force it.  Also, what if `task.get("returnCode")` actually is `None`? If that happens you would need to return something that is a number otherwise it will fail when the CLI attempts to marshall the JSON.

How about:

```
exit_str = (
                    "" if task.get("returnCode") is None else task.get("returnCode")
                )
                try:
                    exit_code = int(exit_str)
                except (TypeError, ValueError):
                    self.logger.info(
                        f"exit code of '{int(task.get('returnCode'))}' is not an integer, will return 1 instead"
                    )
                    exit_code = 1

                log = {
                    "name": task_name + "|" + task.get("jobId", "XXXXX"),
                    "cmd": [task.get("commandLine")],
                    "start_time": task.get("start"),
                    "end_time": task.get("end"),
                    "stdout": task.get("stdout"),
                    "stderr": task.get("stderr"),
                    "exit_code": exit_code,
                }
                task_logs.append(log)
```

markjschreiber (changes requested):

packages/wes_adapter/amazon_genomics/wes/adapters/CromwellWESAdapter.py
OhCloud: We handle safely handle the "Null" use case on the Golang CLI side so we dont need to perform the check. You can take a look at that [here](https://cromwell.readthedocs.io/en/stable/api/RESTAPI/#definitions) (under CallMetadata - returnCode).

markjschreiber (approved):

elliot-smith (approved):

Issue #204, opened 2021-12-02
Change the Nextflow head node instance size and type


snamburi3:
**Use Case**

The Nextflow head node is currently run on c5.large instance. The head node sometimes crashes due to memory issues or other reasons.

**Proposed Solution**

Ability to change the instance size and type of the nextflow head node

multimeric:
Okay I've found a better hack:

1. Find the head node job definition. On [this page](https://console.aws.amazon.com/batch/home#job-definition) it should appear as something like `NextflowEngineNexflowHe-6710f13f7c41ae6`
2. Select the radio button next to the job definition in the list and click Create Revision
3. Adjust the system resources as necessary, and then save
4. Note the ARN of the new job definition: this should be the same as the previous one, but with `:2` at the end
5. Look for the WES adapter lambda. On [this page](https://console.aws.amazon.com/lambda/home#/functions) it will be the function with the name `Agc-Context-PIPELINE-USERNAME-NextflowWesAdapterLambda-ID` where `PIPELINE` etc are variables for your own workflow
6. Open the function, click Configuration, and edit the `JOB_DEFINITION` environment variable to the new ARN noted in step 4
7. Save, and wait for about 5 minutes for the lambda to update
8. Run a new workflow, and it should now be using the updated system resources

markjschreiber:
The instance size is chosen by AWS Batch based on the CPU and RAM requested in the job definition. It would be reasonable to be allowed to specify the CPU and RAM of an engine if the defaults cause a larger workflow to fail. Some engines (such as miniwdl) let you do this via command line flags and adjust their own job definition however not all do so adding an optional config would be a useful feature. We have put it into our backlog.

PR #413, opened 2022-04-20
Feat: Adding the option to provide cpu count and memory size for the …

josuebc:
Adding the option to provide cpu count and memory size for the engine via the yaml project file.

Issue: https://github.com/aws/amazon-genomics-cli/issues/204

**Description of Changes**
I added the ability to change the cpu count and memory limit of the engine within a specific context. 

These was accomplish by adding two new variables that get included on the call to the backend, they get included with all existing variables.

In the backend these new variables will only be used by the engines logic when setting up the workflow via "agc context deploy --context <Context Name>" 

engines in context for these changes are: Cromwell, Toil, miniwdl, nextflow and snakemake .

Unit Test for the changes were verified all pass.
Exiting unit test affected by the changes were update and validated.

Lastly I build the CLI and ran multiple variation for all engines how the end user would update the project file and verified all got reflected as expected in the workflows   

agc account activate
agc context deploy --context spotContext
agc workflow run hello --context spotContext

**Checklist**

- [X] I have added/updated unit tests that prove my fix is effective or that my feature works
- [ ] I have linted my code before raising the PR

----

### DIFF
--- a/site/content/en/docs/Concepts/contexts.md
+++ b/site/content/en/docs/Concepts/contexts.md
@@ -116,7 +116,7 @@ contexts:
 | Toil      | 2     | 16384              |
 
 
-You may optionally specify the number of vCpus and memory used in a context job, vCpu count for engines Cromwell or Toil would be converted to cpu units, 1024 per each vCPU.
+You may optionally specify the number of vCpus and memory used in a context definition, vCpu count for server engines such as Cromwell or Toil would be converted to cpu units, 1024 per each vCPU.
 
 *note:* if these are not provided the defaults on the table above would be used.
 ```yaml


examples/demo-nextflow-project/agc-project.yaml
```diff
@@ -26,3 +26,6 @@ contexts:
     engines:
       - type: nextflow
         engine: nextflow
+        resourceRequirements:
+            vcpus: 1
```
markjschreiber, line 30: Nit - this may as well be 2 as the smallest instance size will have 2CPU

markjschreiber (dismissed):

packages/cdk/lib/env/context-app-parameters.ts
```diff
@@ -87,6 +87,14 @@ export class ContextAppParameters {
    * The types of EC2 instances that may be launched in the compute environment.
    */
   public readonly instanceTypes?: InstanceType[];
+  /**
+   * The maximum number of Amazon EC2 vCPUs that an environment can reach for the nextflow engine.
+   */
+  public readonly vCpus?: number;
+  /**
+   * The maximum memory limit that an environment can reach for the nextflow engine.
```
IllyaYalovyy, line 95: Is this NextFlow specific? How it will affect other supported engines? Have we tested this change with other engines?

josuebc: Yes this is NextFlow specific.

examples/demo-nextflow-project/agc-project.yaml
josuebc: Thanks, I updated it to 2 as suggested.

markjschreiber (changes requested):
The JSON schema and binding would allow you to provide these parameters for any engine which creates the impression you can do it for all. The docs also don't say this is only for Nextflow.

Is there a reason that the approach cannot be generalized to all engines? Or at least those engines that run as Batch Jobs (miniwdl, nextflow and snakemake).

examples/demo-wdl-project/agc-project.yaml
```diff
@@ -32,6 +32,9 @@ contexts:
     engines:
       - type: wdl
         engine: miniwdl
+        resourceRequirements:
```
markjschreiber, line 35: Are engine `resourceRequirements` a required property now? If they are not then we probably don't need them in the example `agc.project-yaml` files unless we know these values are needed to run the example workflows

packages/cdk/lib/constructs/engines/engine-job-definition.ts
```diff
@@ -19,8 +19,8 @@ export class EngineJobDefinition extends JobDefinition {
           AWS_METADATA_SERVICE_NUM_ATTEMPTS: "10",
           ...props.container.environment,
         },
-        memoryLimitMiB: props.container.memoryLimitMiB || 2048,
-        vcpus: props.container.vcpus || 1,
+        memoryLimitMiB: props.container.memoryLimitMiB || 4096,
```
markjschreiber, line 22: We should actually probably keep this at 1 vcpu and 4096 RAM. When I suggest 2 vCPU I had forgotten that these are Fargate jobs so it is possible for it to be 1. Sorry about that.

markjschreiber (commented):
Would it be possible to extend this approach to include Cromwell and Toil? Although they run as ECS tasks rather than Batch jobs they have similarly configurable resource requirements. You would need to make sure that the minimum values are what they are now.

The ECS services are created by `packages/cdk/lib/constructs/secure-service.ts` and the `cpu` and `memory` values are in the `SecureServiceProps`

examples/demo-wdl-project/agc-project.yaml
josuebc: these are not a required property, I remove them from the example project files.

packages/cdk/lib/constructs/engines/engine-job-definition.ts
josuebc: np, changing it back.

site/content/en/docs/Concepts/contexts.md
```diff
@@ -105,7 +105,31 @@ contexts:
...
+| Engine    | vCpus | Memory value (MiB) |
+|-----------|-------|--------------------|
+| Cromwell  | 2     | 16384              |
+| Nextflow  | 1     | 2048               |
+| miniwdl   | 2     | 4096               |
+| Snakemake | 2     | 4096               |
+| Toil      | 2     | 16384              |
+
+
+You may optionally specify the number of vCpus and memory used in a context job, vCpu count for engines Cromwell or Toil would be converted to cpu units, 1024 per each vCPU.
```
markjschreiber, line 119:
```suggestion
You may optionally specify the number of vCpus and memory used in a context definition, vCpu count for server engines such as Cromwell or Toil would be converted to cpu units, 1024 per each vCPU.
```

markjschreiber (dismissed):
One minor suggestion in the docs but otherwise great.

markjschreiber (approved):

packages/cdk/lib/constructs/engines/snakemake/snakemake-engine.ts
```diff
@@ -41,7 +42,8 @@ export class SnakemakeEngine extends Engine {
       logGroup: this.logGroup,
       platformCapabilities: [PlatformCapabilities.FARGATE],
       container: {
-        memoryLimitMiB: this.engineMemoryMiB,
+        vcpus: props.vcpus || 1,
```
IllyaYalovyy, line 45: can you create a constant with a meaningful name for this magic number? 

```
const defaultVCPUs = 1;
```

packages/cdk/lib/constructs/engines/snakemake/snakemake-engine.ts
```diff
@@ -41,7 +42,8 @@ export class SnakemakeEngine extends Engine {
...
+        memoryLimitMiB: props.engineMemoryMiB || 4096,
```
IllyaYalovyy, line 46: can you create a constant with a meaningful name for this magic number? 

```
const defaultMemoryMiB = 4096;
```

packages/cdk/lib/env/context-app-parameters.ts
IllyaYalovyy: Should you update comments than?

packages/cdk/lib/stacks/engines/cromwell-engine-construct.ts
```diff
@@ -57,7 +58,7 @@ export class CromwellEngineConstruct extends EngineConstruct {
     });
 
     // TODO: Move log group creation into service construct and make it a property
-    this.engine = this.getEngineServiceDefinition(props.vpc, props.subnets, engineContainer, this.engineLogGroup);
+    this.engine = this.getEngineServiceDefinition(props.vpc, props.subnets, engineContainer, this.engineLogGroup, params);
```
IllyaYalovyy, line 61: I think you just cut in between the line of code and the comment that corresponds to that line of code. Now this TODO comment sounds out of context. Could you please rearrange code, so the comment stays close to the related line?

packages/cdk/lib/stacks/engines/cromwell-engine-construct.ts
```diff
@@ -99,8 +106,8 @@ export class CromwellEngineConstruct extends EngineConstruct {
     });
     const definition = new FargateTaskDefinition(this, "EngineTaskDef", {
       taskRole: this.engineRole,
-      cpu: serviceContainer.cpu,
-      memoryLimitMiB: serviceContainer.memoryLimitMiB,
+      cpu: (params.vCpus ? params.vCpus * 1024 : undefined) || serviceContainer.cpu,
```
IllyaYalovyy, line 109: What is the meaning of this `1024` magic number? can you use a constant with a meaningful name instead? I think it will improve readability and will serve self-documenting purpose.

packages/cdk/lib/stacks/engines/cromwell-engine-construct.ts
```diff
@@ -112,8 +119,8 @@ export class CromwellEngineConstruct extends EngineConstruct {
     });
 
     const container = definition.addContainer(serviceContainer.serviceName, {
-      cpu: serviceContainer.cpu,
-      memoryLimitMiB: serviceContainer.memoryLimitMiB,
+      cpu: (params.vCpus ? params.vCpus * 1024 : undefined) || serviceContainer.cpu,
```
IllyaYalovyy, line 122: ditto. By the way, may be we should consider avoiding this duplicated code block?

packages/cdk/lib/stacks/engines/toil-engine-construct.ts
```diff
@@ -74,17 +75,23 @@ export class ToilEngineConstruct extends EngineConstruct {
...
+    serviceContainer: ServiceContainer,
+    logGroup: ILogGroup,
+    params: ContextAppParameters
+  ) {
     const id = "Engine";
     const definition = new FargateTaskDefinition(this, "EngineTaskDef", {
       taskRole: this.engineRole,
-      cpu: serviceContainer.cpu,
-      memoryLimitMiB: serviceContainer.memoryLimitMiB,
+      cpu: (params.vCpus ? params.vCpus * 1024 : undefined) || serviceContainer.cpu,
```
IllyaYalovyy, line 88: Here we go again... I think if some number has to be repeated more than once, it should be moved to a constant.

packages/cdk/lib/stacks/engines/toil-engine-construct.ts
```diff
@@ -74,17 +75,23 @@ export class ToilEngineConstruct extends EngineConstruct {
...
+      memoryLimitMiB: params.memoryLimitMiB || serviceContainer.memoryLimitMiB,
     });
 
     definition.addContainer(serviceContainer.serviceName, {
-      cpu: serviceContainer.cpu,
-      memoryLimitMiB: serviceContainer.memoryLimitMiB,
+      cpu: (params.vCpus ? params.vCpus * 1024 : undefined) || serviceContainer.cpu,
```
IllyaYalovyy, line 93: and one more time...

packages/cli/internal/pkg/cli/spec/context.go
```diff
@@ -10,10 +10,17 @@ type Filesystem struct {
...
-	Filesystem Filesystem `yaml:"filesystem,omitempty"`
+	Type                 string              `yaml:"type"`
+	Engine               string              `yaml:"engine"`
+	ResourceRequirements ResourceRequirement `yaml:"resourceRequirements,omitempty"`
+	Filesystem           Filesystem          `yaml:"filesystem,omitempty"`
 }
+
+type ResourceRequirement struct {
+	VCpus          int `yaml:"vcpus,omitempty"`
+	MemoryLimitMiB int `yaml:"memoryLimit,omitempty"`
```
IllyaYalovyy, line 21: I don't think it will work as you might expect. "omitempty" means not to include values that absent. In your case values **always** present. 

Here is the test: 
https://go.dev/play/p/WB4AzY1JhTT

IllyaYalovyy: I would suggest a unit test for that.

packages/cli/internal/pkg/cli/spec/project_test.go
```diff
@@ -122,6 +126,10 @@ schemaVersion: 0
 							{
 								Type:   "nextflow",
 								Engine: "nextflow",
+								ResourceRequirements: ResourceRequirement{
+									VCpus:          2,
```
IllyaYalovyy, line 130: Lets add tests with partial configuration. e.g. either VCpus or MemoryLimitMiB is set, but not both.

IllyaYalovyy (changes requested):

PR #422, opened 2022-05-03
chore: Rebase of adams pr

markjschreiber:
Re based on origin/Main

**Checklist**

----


packages/cdk/lib/roles/policies/toil-batch-policy.ts
```diff
@@ -0,0 +1,30 @@
+import { PolicyStatement, Effect } from "aws-cdk-lib/aws-iam";
+import { CromwellBatchPolicy } from "./cromwell-batch-policy";
+
+export interface ToilBatchPolicyProps {
+  jobQueueArn: string;
+  toilJobArnPattern: string;
+}
+
+export class ToilBatchPolicy extends CromwellBatchPolicy {
```
wleepang, line 9: Noting we should generalize this further - e.g. a ServerBatchPolicy that both Cromwell and Toil extend.

packages/cdk/lib/roles/toil-job-role.ts
```diff
@@ -0,0 +1,126 @@
...
+    // becomes able to just use the one AGC-provided bucket.
+    const jobStoreS3ArnPattern = Arn.format(
+      {
+        // Note that regions and account IDs aren't allowed in S3 ARNs.
+        // But the formatter requires them to be passed, and wants an empty
+        // string if we don't actually want them filled in.
+        account: "",
+        region: "",
+        partition: Aws.PARTITION,
+        resource: "toil-*",
```
wleepang, line 47: Note: this is a separate bucket from the default bucket used by agc.

packages/engines/toil/README.md
```diff
@@ -0,0 +1,51 @@
+## Toil AWS Mirror
+
+A Toil mono-container WES server for use with Amazon Genomics CLI.
+
+### Building the Container Manually
+
+Go to this directory and run:
+
+```bash
+docker build . -f Dockerfile -t adamnovak/toil-agc
```
wleepang, line 10: is this correct container image uri?

wleepang (approved):

packages/engines/toil/README.md
adamnovak: That's not a tag I own on Docker Hub, but you can tag as that. The build procedure just needs an intermediate tag to avoid you always needing to use the long ECR repo tag.

I have some improvements to the Docker build procedure in https://github.com/adamnovak/amazon-genomics-cli/commit/26dc5add75f8091fcfbf012bfcc2eafc060548e7 where I fix a mistake in the example commands for the push, and also stop using grep to find stuff for local testing. I rebased that over onto here in c017154ee13261c3f4b93954dcad6a9ff7a41b37.

IllyaYalovyy (approved):
These seem to be work in progress, but I think we can iterate faster with smaller PRs

adamnovak:
According to my testing, this needs be5e01becd1b0eead506271fcef5c71b41ac6125, granting `s3:GetBucketLocation`, to actually be able to run workflows.

It could use d979cacb14a6beb160f9ce6bff1fc4d66fc9bc1e to remove a few permissions that don't exist, and c017154ee13261c3f4b93954dcad6a9ff7a41b37 to make sure you can actually follow along with the Toil engine README.

Issue #430, opened 2022-05-05
Cannot access Toil "adapter" logs


markjschreiber:
**Describe the Bug**

The API Gateway and Wes lambda logs are not available from the CLI for Toil contexts

**Steps to Reproduce**

```shell
agc logs adapter -c myContext
2022-05-05T10:58:27-04:00 𝒊  Showing adapter logs for 'myContext'
2022-05-05T10:58:30-04:00 ✘   error="operation error CloudWatch Logs: FilterLogEvents, https response error StatusCode: 400, RequestID: 6fe4a223-406e-44a6-8006-947adccf2eb1, InvalidParameterException: 2 validation errors detected: Value '' at 'logGroupName' failed to satisfy constraint: Member must have length greater than or equal to 1; Value '' at 'logGroupName' failed to satisfy constraint: Member must satisfy regular expression pattern: [\\.\\-_/#A-Za-z0-9]+"
```

**Additional Context**

Error suggests that AGC is not assembling the correct log group name

Operating System:
AGC Version: Git HEAD
Was AGC setup with a custom bucket: No
Was AGC setup with a custom VPC: No

adamnovak:
The answer here is probably to detect that the log group name is empty (or that the engine is Toil), and tell the user that there are no adapter logs to be got.

PR #431, opened 2022-05-05
fix: make agc logs adapter handle Toil

adamnovak:
Issue #, if available: #430

**Description of Changes**

This changes `agc logs adapter` so that it doesn't try to fetch anything for a Toil context, and instead tells the user that there's no adapter.

I added a test, and I also tried them on the command line:

```
[anovak@courtyard demo-cwl-project]$ agc logs adapter -c myContext
Error: Context does not use an adapter because it is using the Toil engine
```

**Checklist**

----

### DIFF
--- a/packages/cli/internal/pkg/cli/logs_adapter.go
+++ b/packages/cli/internal/pkg/cli/logs_adapter.go
@@ -4,9 +4,12 @@
 package cli
 
 import (
+	"fmt"
+
 	"github.com/aws/amazon-genomics-cli/internal/pkg/aws"
 	"github.com/aws/amazon-genomics-cli/internal/pkg/cli/clierror"
 	"github.com/aws/amazon-genomics-cli/internal/pkg/cli/context"
+	"github.com/aws/amazon-genomics-cli/internal/pkg/constants"
 	"github.com/rs/zerolog/log"
 	"github.com/spf13/cobra"
 )
@@ -33,6 +36,22 @@ func newLogsAdapterOpts(vars logsAdapterVars) (*logsAdapterOpts, error) {
 func (o *logsAdapterOpts) Validate() error {
 	if err := o.validateFlags(); err != nil {
 		return err
+	}
+
+	// It's convenient for the tests to be able to not actually have the ctxManager set.
+	if o.ctxManager != nil {
+		ctxMap, err := o.ctxManager.List()
+		if err != nil {
+			return err
+		}
+
+		summary := ctxMap[o.contextName]
+		engine := summary.Engines[0].Engine
+		if engine == constants.TOIL {
+			// The Toil engine doesn't use an adapter, so we shouldn't let the user
+			// ask for adapter logs.
+			return fmt.Errorf("Context does not use an adapter because it is using the Toil engine")
+		}
 	}
 
 	return o.parseTime(o.logsSharedVars)
--- a/packages/cli/internal/pkg/cli/logs_adapter_test.go
+++ b/packages/cli/internal/pkg/cli/logs_adapter_test.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/aws/amazon-genomics-cli/internal/pkg/aws/cwl"
 	"github.com/aws/amazon-genomics-cli/internal/pkg/cli/context"
+	"github.com/aws/amazon-genomics-cli/internal/pkg/cli/spec"
 	awsmocks "github.com/aws/amazon-genomics-cli/internal/pkg/mocks/aws"
 	contextmocks "github.com/aws/amazon-genomics-cli/internal/pkg/mocks/context"
 	"github.com/golang/mock/gomock"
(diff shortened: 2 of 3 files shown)


packages/cli/internal/pkg/cli/logs_adapter.go
```diff
@@ -35,6 +38,22 @@ func (o *logsAdapterOpts) Validate() error {
 		return err
 	}
 
+	// It's convenient for the tests to be able to not actually have the ctxManager set.
```
IllyaYalovyy, line 41: I think this comment doesn't add much information, and can be removed in order to keep code clean.

packages/cli/internal/pkg/cli/logs_adapter.go
```diff
@@ -35,6 +38,22 @@ func (o *logsAdapterOpts) Validate() error {
...
+	if o.ctxManager != nil {
+		ctxMap, err := o.ctxManager.List()
+		if err != nil {
+			return err
+		}
+
+		summary := ctxMap[o.contextName]
+		engine := summary.Engines[0].Engine
+		if engine == constants.TOIL {
```
IllyaYalovyy, line 50: I think we can avoid explaining code using comments if we use self documenting code instead. It will have additional benefit of forward looking design. In this case the better option might be to introduce a list of engines with native WES API support.

```go
var HasNativeWES = map[sting]bool{constants.TOIL: true}
...
if HasNativeWES[engine] {
  return fmt.Errorf("'%s' doesn't use an adapter, because it supports WES API natively", engine)
}



```

IllyaYalovyy (changes requested):

packages/cli/internal/pkg/cli/logs_adapter.go
adamnovak: Do we need to explain somewhere else why we need to support `ctxManager` being `nil` when it isn't ever in normal operation? Or do we just let the tests constrain future developers who might think it was safe to remove the check?

IllyaYalovyy: It is a very good point.
I think that the fact, that we have to "explain" something is a red flag. It meas 1) code is not clear and self explanatory 2) we are doing something wrong.

There are 2 questions here.
1) is `nil` a valid state of a context manager? It is not. It means that ignoring that fact is not a correct behavior of the "validate" method, because it can produce unexpected and hard to troubleshoot problems in production
2) Should business logic be tailored to unit test? No, it should not. It's effectively makes unit testing less efficient, and add complexity to our main logic. 

My recommendation here is to remove this check, and make sure we properly populate ctxManager in tests. In this case we:
1) don't mix test logic with business logic
2) do not introduce any implicit dependency that has to be explained in a comment or documentation
3) make sure we are testing actual code flow in our unit tests.

packages/cli/internal/pkg/cli/logs_adapter.go
```diff
@@ -35,6 +38,22 @@ func (o *logsAdapterOpts) Validate() error {
 		return err
 	}
 
+	// It's convenient for the tests to be able to not actually have the ctxManager set.
+	if o.ctxManager != nil {
```
IllyaYalovyy, line 42: It is better to keep method structure flat. Here is an example:
```go
if o.ctxManager == nil {
  return ...
}
// do something with o.ctxManager

```

IllyaYalovyy (changes requested):

packages/cli/internal/pkg/cli/logs_adapter.go
adamnovak: I've set up a map for this, and plugged it in to the `environment.go` stuff that already needed this information.

adamnovak:
OK, I set this up to use a `UsesWesAdapter` map, but I had to tie it in with the information about what engines need the adapter that was already in `environment/environment.go` to avoid duplicating this information. And to do that I had to standardize on the lowercase names for everything from `constants/engines.go` instead of the upper-case image keys that had been in use in some places.

I think I have it all set up in a sensible way now; I've thrown out the dependency resolution logic I had before in favor of just checking the do-we-need-an-adapter flag. The only downside is that `UsesWesAdapter` is coming from `environment` instead of some kind of `engines` package. Should it maybe move?

packages/cli/internal/pkg/cli/logs_adapter.go
adamnovak: OK, I've set up the tests for adapter logs to all have a stub context manager, that will return contexts if requested but that won't insist on being called. I think that's the contract we want fulfilled here.

Then in `Validate()` I'm making sure that the context manager is set.

So the tests now pass and should keep passing even if the context manager checks move relative to the other validation checks, and tests start/stop getting the context.

packages/cli/internal/pkg/environment/environment.go
```diff
@@ -1,78 +1,72 @@
...
 import (
+	"fmt"
 	"os"
+	"strings"
 
 	"github.com/aws/amazon-genomics-cli/internal/pkg/aws/ecr"
+	"github.com/aws/amazon-genomics-cli/internal/pkg/constants"
 )
 
+var AllEngines = []string{
```
IllyaYalovyy, line 12: I really like that you added this list. I agree that now, it makes sense to move engine related code to a separate package. It doesn't have to be this PR though. We can do that later. Another thing that we may want to do, is a quick unit test, that makes sure, that AllEngine list is in sync with all the maps. To catch cases when we added something to AllEngines, but forget to add it to UsesWesAdapter or another way around.

waliag87 (approved):

Issue #458, opened 2022-05-20
Add account activate flag for using private API Gateway


markjschreiber:
[//]: #  (Before raising the feature request, please check to see if an existing feature request already exists.)

**Description**

Some users want to restrict API Gateway endpoints for all contexts to be private and connect through the VPC Gateway.

**Use Case**

Secures access to make API Gateway endpoints only accessible through the VPC Endpoint for API gateway.

**Proposed Solution**

Similar to the `--vpc`, `--subnets` and `--ami` flags, this would be applied at `agc account activate` and recorded as an SSM parameter that would be used for all `agc context create` commands as a value that would be passed through to the CDK that creates the API Gateway for the context.

The new flag should only be used if the `--vpc` is set AND the `--vpc` referenced needs to have a VPC Endpoint of the `com.amazonaws.<region>.execute-api` type. Not sure if it is possible to look this up or if the ID of this would have to be provided with the flag.

**Other information**

Potential gist https://gist.github.com/skorfmann/6941326b2dd75f52cb67e1853c5f8601

markjschreiber:
On the CDK side we would need to change `packages/cdk/lib/constructs/api-proxy.ts` line 61 to:

`endpointTypes: [EndpointType.PRIVATE],`

It would technically be possible to have a private and regional gateway although then you would need to specify this and AGC would need to figure out which is which.

PR #507, opened 2022-08-18
feat: support for PRIVATE API Gateway endpoints

markjschreiber:
Issue #, if available:
#458 

**Description of Changes**

Added CLI support for requesting the use of `PRIVATE` ApiGateway endpoints when activating an account so that all contexts subsequently deployed will use that type of gateway.

`make && make init`

* activate account with use of private endpoint or private endpoint with pre-existing endpoint ID in the specified VPC. 
* Verified that the configuration is stored in SSM Parameter Store
* Verify that deploying contexts adopt these settings and workflows can be run.

**Checklist**

----


lingjiekong (approved):
LGTM. BTW, do we need to update the corresponding document regarding now there is an option to configure endpoint type and it is default to `REGIONAL`.

markjschreiber:
> LGTM. BTW, do we need to update the corresponding document regarding now there is an option to configure endpoint type and it is default to `REGIONAL`.

Docs are already updated in this PR

Issue #393, opened 2022-04-07
ARM support for install.sh


bollig:
[//]: #  (Before raising the feature request, please check to see if an existing feature request already exists.)

**Description** attempted to install on t4g instance type and ran into an unsupported arch error. Aarch64 would be great for cost-effective bastion/cli. 

[//]: #  (A short description of the feature you are proposing.)

**Use Case**

[//]: #  (Why do you need this feature?)

**Proposed Solution**

[//]: #  (Please include prototype/workaround/sketch/reference implementation.)

**Other information**

[//]: #  (detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. associated pull-request, stackoverflow, slack, etc)

wleepang:
Proposed solution - add an additional compile target here:
https://github.com/aws/amazon-genomics-cli/blob/6eaa773b86aab2f98c174da756bc912ecf287ad7/packages/cli/Makefile#L32-L39

m3talsmith:
@wleepang I ended up wrapping your proposal in with some adjustments for Darwin and Linux ARM CLI build and release support in #569

m3talsmith:
@bollig ARM CLI build support, including `install.sh`, has been merged into main. I would be interested in your feedback, should you hit any additional issues related.

PR #569, opened 2022-11-17
feat: Add ARM support for CLI build

m3talsmith:
## Summary

This commit focuses on two related `agc` CLI build and release issues:

- Apple ARM support in V8
- ARM build and release support

All of these issues are focused on developing the CLI and nothing else. Details follow below.

### Apple ARM support in V8

The current version of NodeJS set in `.nvmrc` has a version of V8 that does not support Apple's ARM family of processors. This caused issues where following the [README](https://github.com/aws/amazon-genomics-cli#readme) for building the CLI, failed at the [NVM step](https://github.com/aws/amazon-genomics-cli#node). Support for Apple ARM was back ported to NodeJs v14 LTS in 14.17.0.

Additionaly, NodeJs compilation would fail if you had Python 1.10 or greater installed on your Mac.

#### Fixes

- Uses newer V8 support for Python 3.10+
- Uses newer V8 support of Apple ARM processor family
- Includes various NodeJS v14 LTS bug and security fixes

### ARM build and release support

The current Make compile targets made assumptions about what the Go architecture and output would be; these were amd64 centric.

As well, there was a subtle bug in the release packaging that assumed jq, as a dependency, was installed; which may not be true by default.

#### Fixes

- Add ARM compile targets for Darwin and Linux
- Include ARM compile targets in release packaging
- Exit release packing early if missing dependency jq

## This closes issues

- #393 

### Apple ARM support in V8

1. Follow the [README One Time Setup steps](https://github.com/aws/amazon-genomics-cli#one-time-setup) on an Apple M1 or higher family of processors
1. Verify that NodeJs doesn't error

### ARM build and release support

1. Uninstall `jq` if installed
2. Run `make release`
  1. Expect an error and an exit 1 status from `package-release.sh`
3. Install `jq`
4. Run `make release`
  1. Expect no error and an exit 0 status from `package-release.sh`

## Checklist

This change is local development and only changes build and release scripts. No unit testing is available for it.

----


scripts/package-release.sh
```diff
@@ -3,6 +3,12 @@
 set -eo pipefail
 
 RELEASE_DIR="dist/amazon-genomics-cli"
+JQ=$(which jq)
```
markjschreiber, line 6: The `which` command is not found on some linux distros. A more POSIX compliant and portable way to check for the presence of an application is:

```
if ! command -v <the_command> &> /dev/null
then
    echo "<the_command> could not be found"
    exit
fi
```

markjschreiber (changes requested):

.nvmrc
```diff
@@ -1 +1 @@
-14.15.0
+14.21.1
```
markjschreiber, line 1: Changing this will likely require changes in the github actions. Is there a strong reason to change?

scripts/package-release.sh
m3talsmith: @markjschreiber good point. It's better to stay as close to POSIX as possible.

.nvmrc
m3talsmith: I'll make the change to the github actions. This is the change that gets highlighted first in [my notes]() because it pulls in the NodeJs version which contains ARM support via a V8 version:

> Apple ARM support in V8
>
> The current version of NodeJS set in .nvmrc has a version of V8 that does not support Apple's ARM family of processors. This caused issues where following the [README](https://github.com/aws/amazon-genomics-cli#readme) for building the CLI, failed at the [NVM step](https://github.com/aws/amazon-genomics-cli#node). Support for Apple ARM was back ported to NodeJs v14 LTS in 14.17.0.
>
> Additionaly, NodeJs compilation would fail if you had Python 1.10 or greater installed on your Mac.
>
> Fixes
>
> - Uses newer V8 support for Python 3.10+
> - Uses newer V8 support of Apple ARM processor family
> - Includes various NodeJS v14 LTS bug and security fixes

markjschreiber (changes requested):

scripts/cli/install.sh
```diff
@@ -6,11 +6,30 @@ USER_BIN_DIR="$HOME/bin"
...
+
+    if [[ -z "$archKind" ]]; then
+        eval $filename="agc"
+    else
+        eval $fileName="$filename-$archKind"
+    fi
+
     if [[ "$OSTYPE" == "linux-gnu"* ]]; then
-        eval $1="agc-amd64"
+        eval $1=filename
```
mattcarp12, line 30: I think this change is breaking the script.

m3talsmith: @mattcarp12 what are you seeing here? Is there an error that's getting output? Is the script exiting? Is the binary being misnamed?

mattcarp12: Sorry, yes when I run this on linux I get `cp: cannot stat ‘/home/carpedma/amazon-genomics-cli/dist/amazon-genomics-cli/filename’: No such file or directory`

m3talsmith: This is definitely a misnamed variable. I'm fixing that, but also making sure nothing else problematic is happening.

m3talsmith: Ok, fix is in a PR since this one is closed: https://github.com/aws/amazon-genomics-cli/pull/575

PR #571, opened 2022-11-18
feat: Add resource tags that indicate engine used by context

mattcarp12:
Issue #, if available:

**Description of Changes**

**Checklist**

----


packages/cdk/lib/env/context-app-parameters.ts
```diff
@@ -212,4 +218,28 @@ export class ContextAppParameters {
     }
     return defFilesystem;
   }
+
+  public getEngineType(): string {
+    let engineType: string;
+    switch (this.engineName) {
```
vvalleru, line 224: Should you ignore case before checking equivalency?

packages/cdk/lib/env/context-app-parameters.ts
```diff
@@ -212,4 +218,28 @@ export class ContextAppParameters {
...
+        engineType = "wdl";
+        break;
+      case "snakemake":
+        engineType = "snakemake";
+        break;
+      case "toil":
+        engineType = "cwl";
+        break;
+      default:
+        throw Error(`Engine '${this.engineName}' is not supported`);
```
vvalleru, line 241: Does it need to be blocking? Just a log with a warn or a “unknown” value should be enough. Thoughts?

packages/cdk/lib/env/context-app-parameters.ts
mattcarp12: Here I was following the pattern of `getDefaultFilesystem()` above, which does not ignore case. Do you suggest I change both to ignore case?

packages/cdk/lib/env/context-app-parameters.ts
mattcarp12: Here just following the pattern of `getDefaultFilesystem()` above, which throws the error. For me, I think it's fine to leave as is.

vvalleru: Tagging the resources is for our analytics and not a "customer use case". So, I we should not block a customer use case (stack deployment in this case) when we fail to associate analytical data.

packages/cdk/lib/env/context-app-parameters.ts
vvalleru: Unlike `filesystem`, the engine name is determined from the customer's `agc-project.yml` file. Its good to have some fool proof. If customer uses case in their project files, we should still be able to determine and tag the resources with the right engine metadata.

vvalleru (changes requested):

packages/cdk/lib/env/context-app-parameters.ts
mattcarp12: Good point, will update.

packages/cdk/lib/env/context-app-parameters.ts
mattcarp12: Yes I agree.

vvalleru (approved):

PR #582, opened 2022-12-27
fix: Allow null kms decrypt policy

m3talsmith:
Issue #, if available:

**Description of Changes**

Previous SSE-KMS policy code in the CDK adaptor throws an error if the KMS key is empty, even when it's not needed. This fix uses a method that catches the throw and allows a default null value.

Ran all tests on an AL2 instance.

**Checklist**

----


lingjiekong (approved):
LGTM.

PR #637, opened 2024-02-23
fix: Comment out Lambda return value to avoid nodejs14.x

### LINKED ISSUE #624: Node.js 14
I got a notice from AWS:
"We are ending support for Node.js 14 in AWS Lambda. This follows Node.js 14 End-Of-Life (EOL) reached on April 30, 2023 [1].
 
As described in the Lambda runtime support policy [2], end of support for language runtimes in Lambda happens in two stages. Starting November 27, 2023, Lambda will no longer apply security patches and other updates to the Node.js 14 runtime used by Lambda functions, and functions using Node.js 14 will no longer be eligible for technical support. In addition, you will no longer be able to create new Lambda functions using the Node.js 14 runtime. Starting January 25, 2024, you will no longer be able to update existing functions using the Node.js 14 runtime.
 
We recommend that you upgrade your existing Node.js 14 functions to Node.js 18 before November 27, 2023."

And this shows that the AGC logging is using Node.js 14:
 
   aws lambda list-functions --function-version ALL --region us-east-1 --output text --query "Functions[?Runtime=='nodejs14.x'].FunctionArn"

Returns: 
arn:aws:lambda:us-east-1:<aws acct id>:function:Agc-Context-<context name>-<user>-LogRetentionaae0aa3c5b4d-JWxtx4CDVz1W:$LATEST

Operating System:
AGC Version: v1.6.0
-Standard AGC setup, running a Nextflow pipeline.

bmorrissirromb:
Issue #, if available:
Closes #624 

**Description of Changes**

# Background
Lambda Functions in CDK, by default, do not create Log Groups at CDK deploy time. Log Groups are typically find-or-created when the Lambda executes for the first time.

HOWEVER, there are certain scenarios where the CloudFormation stack needs to know what the Log Group will be. For example, if the Function's log retention needs to be set to a specific value, or if some other resource needs to know the Log Group's properties.

In THESE cases, CDK will generate a `Custom::LogRetention` resource in the background, intended to set retention and return metadata about a Lambda's Log Group. You guessed it, this Custom Resource is itself a Lambda with a runtime of...drumroll...`nodejs14.x` (at least in CDK `2.50.0`). So that's where our problems lies.

# Diving Deep
The question is...why does this get created in this case at all? There's no mention of log retention ANYWHERE in this CDK application. However, careful readers may have picked up on this already: the custom resource also gets created if we're ever reading Log Group metadata. Which we're doing in each of the 4 engine constructs (eg. https://github.com/aws/amazon-genomics-cli/blob/d92ddc0b1e8634480eed743d9bb21fb8350d3111/packages/cdk/lib/stacks/engines/cromwell-engine-construct.ts#L74). That seemingly inoffensive line of code is causing huge deployment issues down the line.

# Solutioning
The next question is what would happen if we just...removed these offending lines of code. And the answer is...pretty much nothing. Setting `this.adapterLogGroup = lambda.logGroup` is effectively unnecessary because the only time that `adapterLogGroup` property is used is as a CfnOutput (which already has error handling if the value is undefined!!).

Therefore, my proposal is to comment out these 4 lines of code. If someone wants to see the Log Group name, they can look it up from the Lambda at Lambda-execute time. I will be submitting a PR for this.

**Validation**

Validated using `cdk synth` locally.

### DIFF
--- a/packages/cdk/lib/stacks/engines/cromwell-engine-construct.ts
+++ b/packages/cdk/lib/stacks/engines/cromwell-engine-construct.ts
@@ -71,10 +71,7 @@ export class CromwellEngineConstruct extends EngineConstruct {
       userId: params.userId,
       engineEndpoint: this.engine.loadBalancer.loadBalancerDnsName,
     });
-    // Referencing the Lambda's logGroup property causes a deprecated
-    // NodeJS14.x custom resource Lambda to be created in the background
-    // Do not reference the Lambda's logGroup or the CFT stack creation will fail
-    // this.adapterLogGroup = lambda.logGroup;
+    this.adapterLogGroup = LogGroup.fromLogGroupName(this, "CromwellAdapterLogGroup", "/aws/lambda/" + lambda.functionName);
 
     this.apiProxy = new ApiProxy(this, {
       apiName: `${params.projectName}${params.contextName}${engineContainer.serviceName}ApiProxy`,
--- a/packages/cdk/lib/stacks/engines/cromwell-engine-construct.ts
+++ b/packages/cdk/lib/stacks/engines/cromwell-engine-construct.ts
@@ -71,7 +71,10 @@ export class CromwellEngineConstruct extends EngineConstruct {
       userId: params.userId,
       engineEndpoint: this.engine.loadBalancer.loadBalancerDnsName,
     });
-    this.adapterLogGroup = lambda.logGroup;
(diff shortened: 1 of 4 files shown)


wleepang:
After some testing:
This fix restores the ability to deploy contexts, but introduces a bug where the CLI client can no longer retrieve adapter logs via:

```
agc logs adapter -c myContext
```

wleepang (approved):

